PR #1, opened 2020-01-07
Add mocked wallets and transactions

valentinacala:
This pr introduces mocked wallets and transactions to display in the Wallet Home of IO app.

image

### DIFF
--- a/payloads/wallet.ts
+++ b/payloads/wallet.ts
@@ -1,0 +1,122 @@
+import { Transaction } from "../generated/definitions/pagopa/Transaction";
+import { range } from "fp-ts/lib/Array";
+import { Wallet } from "../generated/definitions/pagopa/Wallet";
+import { CreditCard } from "../generated/definitions/pagopa/CreditCard";
+import { Psp } from "../generated/definitions/pagopa/Psp";
+import { validatePayload } from "../src/utils/validator";
+import { SessionResponse } from "../generated/definitions/pagopa/SessionResponse";
+import { WalletResponse } from "../generated/definitions/pagopa/WalletResponse";
+import { TransactionResponse } from "../generated/definitions/pagopa/TransactionResponse";
+import { getRandomIntId } from "../src/utils/id";
+
+export const sessionToken: SessionResponse = {
+    data: {
+        sessionToken: '3m3Q2h6e8T5w9t3W8b8y1F4t2m6Q9b8d9N6h1f2H2u0g6E7m9d9E3g7w3T3b5a7I4c4h6U4n2b3Z4p3j8D6p4a5G1c4a8K3o0v8P7e0j6R5i1y2J6d0c7N9i6m0U3j9z'
+    }
+}
+
+export const getWallets = (): WalletResponse => {
+    const validCreditCard: { [key: string]: any } = {
+        id: 1464,
+        holder: "Mario Rossi",
+        pan: "************0111",
+        expireMonth: "05",
+        expireYear: "22",
+        brandLogo:
+          "https://wisp2.pagopa.gov.it/wallet/assets/img/creditcard/carta_mc.png",
+        flag3dsVerified: true
+    };
+  
+    const validAmount: { [key: string]: any } = {
+        currency: "EUR",
+        amount: 1000,
+        decimalDigits: 2
+    };
+  
+    const validPsp: { [key: string]: any } = {
+        id: 43188,
+        idPsp: "idPsp1",
+        businessName: "WHITE bank",
+        paymentType: "CP",
+        idIntermediary: "idIntermediario1",
+        idChannel: "idCanale14",
+        logoPSP:
+          "https://wisp2.pagopa.gov.it/pp-restapi/v2/resources/psp/1578833",
+        serviceLogo:
+          "https://wisp2.pagopa.gov.it/pp-restapi/v2/resources/service/1578833",
+        serviceName: "nomeServizio 10 white",
+        fixedCost: validAmount,
+        appChannel: false,
+        tags: ["MAESTRO"],
+        serviceDescription: "DESCRIZIONE servizio: CP mod1",
+        serviceAvailability: "DISPONIBILITA servizio 24/7",
+        paymentModel: 1,
+        flagStamp: true,
+        idCard: 91,
+        lingua: "IT"
+    };
+
+    enum TypeEnum {
+        "CREDIT_CARD" = "CREDIT_CARD",
+        "BANK_ACCOUNT" = "BANK_ACCOUNT",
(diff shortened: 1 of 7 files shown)


payloads/wallet.ts
```diff
@@ -0,0 +1,122 @@
...
+        pspEditable: true,
+        lastUsage: new Date("2018-08-07T15:50:08Z")
+    };
+  
+    return {
+        data: [WalletBank, WalletCard]
+    } ;   
+}
+
+export const getTransactions = (amount: number): TransactionResponse => {
```
Undermaken:
```suggestion
export const getTransactions = (count: number): TransactionResponse => {
```
I guess count is more appropriate

payloads/wallet.ts
```diff
@@ -0,0 +1,122 @@
...
+    return {
+        data: range(1, amount).map(_ => {
+            return ({
+                accountingStatus: 1,
+                amount: {amount: 20000},
+                created: new Date(2018,10,30,13,12,22,30),
+                description: 'transaction n.' + _,
```
Undermaken:
```suggestion
                description: `transaction n.${_}`,
```
this is because **_** is a number

payloads/wallet.ts
```diff
@@ -0,0 +1,122 @@
...
+                spcNodeStatus: 6,
+                statusMessage: 'statusMessage',
+                success: true,
+                token: 'token',
+                updated: undefined,
+                urlCheckout3ds: 'urlCheckout3ds',
+                urlRedirectPSP: 'urlRedirectPSP'
+            });
+        })
+    }
```
Undermaken: This should return a **TransactionListResponse** and not a **TransactionResponse**
see https://wisp2.pagopa.gov.it/pp-restapi-CD/v2/api-docs
```
[...]
"responses": {
"200": {
"description": "OK",
"schema": {
"$ref": "#/definitions/TransactionListResponse"
}
},
[...]
```
Furthermore to make sure you are creating a valid payload you could hold generated value in a const (ex transactions) and then validate it in return 
`return validatePayload(TransactionListResponse, transactions);`
in this way if your mocked payload is not right compliant (ex you put ) an error will be raised when you compile or start the server (for example I can freely write `create: "hello"` )

payloads/wallet.ts
```diff
@@ -0,0 +1,122 @@
...
+import { TransactionResponse } from "../generated/definitions/pagopa/TransactionResponse";
+import { getRandomIntId } from "../src/utils/id";
+
+export const sessionToken: SessionResponse = {
+    data: {
+        sessionToken: '3m3Q2h6e8T5w9t3W8b8y1F4t2m6Q9b8d9N6h1f2H2u0g6E7m9d9E3g7w3T3b5a7I4c4h6U4n2b3Z4p3j8D6p4a5G1c4a8K3o0v8P7e0j6R5i1y2J6d0c7N9i6m0U3j9z'
+    }
+}
+
+export const getWallets = (): WalletResponse => {
```
Undermaken: same of below here we should return a **WalletListResponse** instead of **WalletResponse**

src/server.ts
```diff
@@ -51,6 +52,21 @@ app.get("/ping", (_, res) => {
 
 export const messages = getMessageWithoutContentList(10, fiscalCode);
 export const services = getServices(10);
+export const wallets = getWallets();
+export const transactions = getTransactions(5);
+
+/** wallet content */
+app.get("/wallet/v1/users/actions/start-session", (_, res) => {
+  res.json(sessionToken)
```
Undermaken:
```suggestion
  res.json(sessionToken);
```

valentinacala: updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9a7b3c5dc219c65fbcbfb87f2e1b29ba2030f538

payloads/wallet.ts
valentinacala: updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9a7b3c5dc219c65fbcbfb87f2e1b29ba2030f538

payloads/wallet.ts
valentinacala: updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9a7b3c5dc219c65fbcbfb87f2e1b29ba2030f538

payloads/wallet.ts
valentinacala: updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9a7b3c5dc219c65fbcbfb87f2e1b29ba2030f538

payloads/wallet.ts
valentinacala: Updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9a7b3c5dc219c65fbcbfb87f2e1b29ba2030f538. 
Anyway, I noticed both the wallet and the transaction types have no required fields

payloads/wallet.ts
```diff
@@ -0,0 +1,126 @@
...
+        data: range(1, count).map(_ => {
+            return ({
+                accountingStatus: 1,
+                amount: {amount: 20000},
+                created: new Date(2018,10,30,13,12,22,30),
+                description: `transaction n.${_}`,
+                error: false,
+                fee: {amount: 1},
+                grandTotal: {amount: 20100},
+                id: getRandomIntId(5),
```
Undermaken: Since the array index increases on each loop you could replace random generation (it could produce a collision anyway) with the index **_** (**_idx_** could be a better name)
So you'll be sure all transactions have an unique id

payloads/wallet.ts
```diff
@@ -0,0 +1,126 @@
...
+        tags: ["MAESTRO"],
+        serviceDescription: "DESCRIZIONE servizio: CP mod1",
+        serviceAvailability: "DISPONIBILITA servizio 24/7",
+        paymentModel: 1,
+        flagStamp: true,
+        idCard: 91,
+        lingua: "IT"
+    };
+
+    enum TypeEnum {
```
Undermaken: you don't need to redefine it
You can import it from `import { TypeEnum } from "../generated/definitions/pagopa/Wallet";`

valentinacala: updated within https://github.com/pagopa/io-dev-api-server/pull/1/commits/9fcc8d8d805d780fe997f6a461e5f3b299750a88

payloads/wallet.ts
valentinacala: updated wihtin https://github.com/pagopa/io-dev-api-server/pull/1/commits/9fcc8d8d805d780fe997f6a461e5f3b299750a88

PR #29, opened 2020-09-18
[#174796762] Adds myportaltoken and bpdToken to getSession call and upgrade api version

CrisTofani:


CrisTofani:
Duplicate of #28

PR #31, opened 2020-11-06
[#175434754] Abort delete userdataprocessing API

debiff:
## Short Description

This PR add the API that allow a user to request the abort of the profile delete

### DIFF
--- a/src/routers/profile.ts
+++ b/src/routers/profile.ts
@@ -116,7 +116,10 @@ installHandler(
   req => {
     const payload = validatePayload(UserDataProcessingChoiceRequest, req.body);
     const choice = payload.choice;
-    if (userChoices[choice] !== undefined) {
+    if (
+      userChoices[choice] !== undefined &&
+      userChoices[choice]?.status !== UserDataProcessingStatusEnum.ABORTED
+    ) {
       return { payload: userChoices[choice] };
     }
     const data: UserDataProcessing = {
--- a/src/routers/profile.ts
+++ b/src/routers/profile.ts
@@ -1,7 +1,10 @@
 import { Router } from "express";
 import { InitializedProfile } from "../../generated/definitions/backend/InitializedProfile";
 import { UserDataProcessing } from "../../generated/definitions/backend/UserDataProcessing";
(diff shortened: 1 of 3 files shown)


src/routers/profile.ts
```diff
@@ -127,6 +127,31 @@ installHandler(
...
 );
 
+installCustomHandler(
+  profileRouter,
+  "delete",
+  addApiV1Prefix("/user-data-processing/:choice"),
+  (req, res) => {
+    const choice = req.params.choice as UserDataProcessingChoiceEnum;
+
+    console.log(choice);
```
Undermaken, line 137:
```suggestion

```

src/routers/profile.ts
```diff
@@ -127,6 +127,31 @@ installHandler(
   }
 );
 
+installCustomHandler(
```
Undermaken, line 130: You can take advantage of codec `UserDataProcessingChoice` and `Option`
This is the refactoring
```typescript
installCustomHandler(
  profileRouter,
  "delete",
  addApiV1Prefix("/user-data-processing/:choice"),
  (req, res) => {
    // try to decode the request param
    const maybeChoice = UserDataProcessingChoice.decode(
      UserDataProcessingChoice
    );
    if (maybeChoice.isLeft()) {
      // the given param is not a valid UserDataProcessingChoice
      // send invalid request
      res.sendStatus(400);
      return;
    }
    const choice = maybeChoice.value;
    // The abort function is managed only for the DELETE
    if (choice === UserDataProcessingChoiceEnum.DOWNLOAD) {
      res.sendStatus(409);
    }

    res.sendStatus(
      fromNullable(userChoices[choice]).fold(409, c =>
        c.status !== UserDataProcessingStatusEnum.PENDING ? 409 : 202
      )
    );
  }
);
```

PS: I didn't test it

Undermaken (changes requested):
Could add some tests about the new implementation [here](https://github.com/pagopa/io-dev-api-server/blob/master/src/routers/__tests__/userDataProcessing.test.ts)?

Undermaken (approved):

PR #37, opened 2021-01-25
[#176443215,#176443259] API for activation and status of CGN

CrisTofani:


src/routers/features/cgn/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      CgnPendingStatus.is(currentCGN)
+        ? res.status(202).json({ id: idActivationCgn })
+        : res.sendStatus(409)
+  );
+});
+
+// Start activation check
+// 200 -> CGN current Status
+// 404 -> no CGN found
+addHandler(bonusVacanze, "get", addPrefix("/status"), (_, res) => {
```
Undermaken, line 55:
```suggestion
addHandler(cgnRouter, "get", addPrefix("/status"), (_, res) => {
```

Undermaken (changes requested):

src/routers/features/cgn/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    currentCGN = {
+      status: ActivatedStatusEnum.ACTIVATED,
+      activation_date: new Date(firstCgnActivationRequestTime),
+      expiration_date: new Date("2050-05-10")
+    };
+    res.status(200).json(currentCGN);
+  } else {
+    res.sendStatus(404);
+  }
+});
```
Undermaken, line 66: Could it be useful export a [reset](https://github.com/pagopa/io-dev-api-server/blob/2e78da43c0a8c232349906d5a8fec4a3d7f8e5fc/src/routers/features/bdp/index.ts#L193) function to restore the initial state?

https://github.com/pagopa/io-dev-api-server/blob/2e78da43c0a8c232349906d5a8fec4a3d7f8e5fc/src/routers/public.ts#L48

Undermaken (changes requested):

src/routers/features/cgn/index.ts
```diff
@@ -0,0 +1,66 @@
+import { Router } from "express";
+import { fromNullable } from "fp-ts/lib/Option";
+import { StatusEnum as ActivatedStatusEnum } from "../../../../generated/definitions/cgn/CgnActivatedStatus";
+import {
+  CgnPendingStatus,
+  StatusEnum as PendingStatusEnum
+} from "../../../../generated/definitions/cgn/CgnPendingStatus";
+// tslint:disable-next-line:no-commented-code
```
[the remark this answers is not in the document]
CrisTofani: Since we have multiple definition of `StatusEnum` I just prefer to leave these imports commented to uncomment in case to receive a different payload

PR #52, opened 2021-07-06
fix 404 url api_backend_specs

hevelius:
api_backend_specs has a wrong url. Changed with right url related to io-backend master branch (maybe could be use release version tag instead).

Also corrected a small spelling mistake ---> api_beckend_specs - api_b[a]ckend_specs


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A mock server to help io-app development https://io.italia.it/",
   "version": "1.0.0",
   "main": "app.js",
-  "api_beckend_specs": "https://raw.githubusercontent.com/pagopa/io-backend/IP-263-update-openapi-spec/api_backend.yaml",
+  "api_backend_specs": "https://raw.githubusercontent.com/pagopa/io-backend/master/api_backend.yaml",
```
Undermaken, line 7: Thank you for your contribution ❤️ 
To keep app and dev-server aligned, could you replace it with the tagged version?

https://github.com/pagopa/io-app/blob/master/package.json#L4

Undermaken (changes requested):

package.json
hevelius: Could make sense to replace also others url pointed to io-backed with tagged version?

Undermaken: yes of course!

hevelius: fix on https://github.com/pagopa/io-dev-api-server/pull/52/commits/3b28f7c7f7614c5f79893ead01752a90fa9ea78d

Undermaken (approved):

PR #56, opened 2021-07-23
[IASV-13] Add Sicilia Vola service

debiff:
This PR add the Sicilia Vola service and the relative CTAs that allow to start the generation and voucher list flows in the app.

### DIFF
--- a/src/payloads/service.ts
+++ b/src/payloads/service.ts
@@ -17,6 +17,12 @@ import {
 } from "../../generated/definitions/backend/ServicePublic";
 import { ServiceScopeEnum } from "../../generated/definitions/backend/ServiceScope";
 import { validatePayload } from "../utils/validator";
+import {
+  frontMatter1CTABonusBpd,
+  frontMatter1CTABonusCgn,
+  frontMatter1CTASiciliaVola,
+  frontMatter2CTA2
+} from "../utils/variables";
 import { IOResponse } from "./response";
 
 export const getService = (serviceId: string): ServicePublic => {
@@ -34,6 +40,22 @@ export const getService = (serviceId: string): ServicePublic => {
   };
   return validatePayload(ServicePublic, service);
 };
+
+export const siciliaVolaServiceId = "serviceSv";
+const siciliaVolaService: ServicePublic = {
+  ...getService(siciliaVolaServiceId),
+  organization_fiscal_code: "18".padStart(11, "0") as OrganizationFiscalCode,
+  organization_name: "Sicilia Vola" as OrganizationName,
+  service_name: "Sicilia Vola" as ServiceName,
+  service_metadata: {
+    scope: ServiceScopeEnum.NATIONAL
+  },
+  version: 1
+};
+
+export const withSiciliaVolaService = (
+  services: readonly ServicePublic[]
+): readonly ServicePublic[] => services.concat(siciliaVolaService);
 
 export const getServices = (count: number): readonly ServicePublic[] => {
   const aggregation = 3;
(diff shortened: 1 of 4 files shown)


src/payloads/service.ts
```diff
@@ -35,6 +41,22 @@ export const getService = (serviceId: string): ServicePublic => {
...
+  organization_fiscal_code: "18".padStart(11, "0") as OrganizationFiscalCode,
+  organization_name: "Sicilia Vola" as OrganizationName,
+  service_name: "Sicilia Vola" as ServiceName,
+  service_metadata: {
+    scope: ServiceScopeEnum.NATIONAL
+  },
+  version: 1
+};
+
+export const withSiciliaVolaService = (
```
Undermaken, line 56: If we add` > 54` services we'll have a clash: there will be one or more servics with the same organization fiscal code of the `sicilia vola`

```typescript
const aggregation = 3;
  // services belong to the same organization for blocks of `aggregation` size
  // tslint:disable-next-line: no-let
  let organizationCount = 0;
  return range(0, count - 1).map(idx => {
    organizationCount =
      idx !== 0 && idx % aggregation === 0
        ? organizationCount + 1
        : organizationCount;
```

It is an edge case but I guess it could be misleading
To make sure `sicilia vola` has a dedicated organization, what about this code?

```typescript
export const withSiciliaVolaService = (
  services: readonly ServicePublic[]
): readonly ServicePublic[] => {
  const organizationsCount = new Set(
    services.map(s => s.organization_fiscal_code)
  ).size;
  return services.concat({
    ...siciliaVolaService,
    organization_fiscal_code: `${organizationsCount + 1}`.padStart(
      11,
      "0"
    ) as OrganizationFiscalCode
  });
};
```

src/payloads/service.ts
```diff
@@ -107,7 +131,11 @@ export const getServiceMetadata = (
     app_android: "https://www.google.com" as NonEmptyString,
     app_ios: "https://www.google.com" as NonEmptyString,
     tos_url: "https://www.tos.com" as NonEmptyString,
-    privacy_url: "https://www.privacy.com" as NonEmptyString
+    privacy_url: "https://www.privacy.com" as NonEmptyString,
+    cta:
+      serviceId.split(".")[0] === siciliaVolaServiceId.toLocaleLowerCase()
```
Undermaken, line 136: We should avoid that a factory function knows about a specific scenario. In that way if we'll add "sicilia Vola2" we have to add extra code into this function and the dev that uses it could have unexpected results

What do you think about a refactoring?
`src/routers/services_metadata.ts`
```typescript
const servicesMetadata = services.map(
  service =>
    getServiceMetadata(service.service_id, visibleServices.payload).payload
);
const servicesMetadataMapping: Record<
  string,
  ServicePublicService_metadata
> = servicesMetadata.reduce((acc, curr, idx) => {
  if (services[idx].service_id === siciliaVolaServiceId) {
    return {
      ...acc,
      [services[idx].service_id.toLowerCase()]: {
        ...curr,
        cta: frontMatter1CTASiciliaVola as NonEmptyString
      }
    };
  }
  return { ...acc, [services[idx].service_id.toLowerCase()]: curr };
}, {});

addHandler(
  servicesMetadataRouter,
  "get",
  addRoutePrefix(`/services/:service_id`),
  (req, res) => {
    const serviceId = req.params.service_id.replace(".json", "").toLowerCase();
    const metadata = servicesMetadataMapping[serviceId as string];
    if (metadata === undefined) {
      res.sendStatus(404);
      return;
    }
    res.json(metadata);
  }
);
```

Undermaken (changes requested):

src/payloads/service.ts
debiff: It works for me! Thank you! Fixed in [38e92a7](https://github.com/pagopa/io-dev-api-server/pull/56/commits/38e92a7cdc63579c033a6f49d9640e60c9ffc2b6)

src/payloads/service.ts
debiff: Yes, it's better! Modified in [389b679](https://github.com/pagopa/io-dev-api-server/pull/56/commits/389b679e75d60aaa6de86ad8a171ae9849c40528)

Undermaken (approved):

PR #79, opened 2021-09-30
[IA-267] Support messages API with pagination

Undermaken:
This PR includes all those changes to support pagination of messages list, as described in this doc
https://pagopa.atlassian.net/wiki/spaces/IC/pages/358909575/Messaggi+-+Paginazione

### DIFF
--- a/src/routers/message.ts
+++ b/src/routers/message.ts
@@ -170,14 +170,31 @@ const createMessages = () => {
 createMessages();
 
 /* helper function to build messages response */
-const getItems = (items: ReadonlyArray<CreatedMessageWithContent>) =>
-  items.map(m => ({
-    id: m.id,
-    fiscal_code: fiscalCode as FiscalCode,
-    created_at: m.created_at,
-    sender_service_id: m.sender_service_id,
-    time_to_live: m.time_to_live
-  }));
+const getItems = (
+  items: ReadonlyArray<CreatedMessageWithContent>,
+  enrichData: boolean
+) => {
+  return items.map(m => {
+    const senderService = services.find(
+      s => s.service_id === m.sender_service_id
+    );
+    const extraData = enrichData
+      ? {
+          service_name: senderService!.service_name,
+          organization_name: senderService!.organization_name,
+          message_title: m.content.subject
+        }
+      : {};
+    return {
+      id: m.id,
+      fiscal_code: fiscalCode as FiscalCode,
+      created_at: m.created_at,
+      sender_service_id: m.sender_service_id,
+      time_to_live: m.time_to_live,
+      ...extraData
+    };
+  });
+};
 
 addHandler(messageRouter, "get", addApiV1Prefix("/messages"), (req, res) => {
   const paginatedQuery = GetMessagesParameters.decode({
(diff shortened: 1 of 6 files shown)


src/routers/message.ts
```diff
@@ -166,17 +167,106 @@ const createMessages = () => {
...
+    minimumId: req.query.minimum_id
+  });
+  if (paginatedQuery.isLeft()) {
+    // bad request
+    res.sendStatus(400);
+    return;
+  }
+
+  const params = paginatedQuery.value;
+  // order messages by creation date (desc)
```
pietro909, line 214: I understand that they are ordered [by ID](https://pagopa.atlassian.net/wiki/spaces/IC/pages/358909575/Messaggi+-+Paginazione#La-soluzione)

Undermaken: Yes, and actually it is!
This implementation is a simplification of what IO backend does. 
IO backend works with a cosmoDB and a storage to retrieve messages.

Messages are immutable so backend orders them by ID and that's ensure they are ordered by creation date:
`ID2 > ID1` -> `creationDate2 > creationDate1`. They use GUID given by cosmosDB to implement this ordering (I guess)

In our scenario we have only the string ID (could represent a number or not) of each message and I think that it could be dangerous trying an ordering using strings. So I point at the expected result and keep it simple: the list should be ordered by creation date

Since your point spawns some doubts about my reasons, I asked to Micheal to confirm them
Schermata 2021-10-01 alle 09 10 06

src/routers/message.ts
```diff
@@ -166,17 +167,106 @@ const createMessages = () => {
...
+    const endIndex = orderedList.findIndex(m => m.id === params.minimumId);
+    if (endIndex <= 0) {
+      res.json({ items: [] });
+      return;
+    }
+    indexes = {
+      startIndex: Math.max(0, endIndex - (1 + params.pageSize!)),
+      endIndex
+    };
+  }
```
pietro909, line 257: I find the large and nested `if/else` block a tad too convoluted. There some exit conditions (the `return`) whithin it that make it hard to follow.

You may initialize `indexes = null` and then match against it down the line.

Example (using `switch` just because I like it 🤓 ) :

```typescript
let indexes = null

switch ([typeof minimumId,typeof maximumId]) {

  // both boundaries
  case ["string", "string"]: {
    // ...more checks
      minimumId.toString()
      indexes = {
        startIndex: startIndex + 1,
        endIndex
      };
  }

  // only lower bound
  case ["string", undefined]: {
    // ...more checks
    indexes = {
      startIndex: Math.max(0, endIndex - (1 + params.pageSize!)),
      endIndex
    };
  }

  // only upper bound
  case [undefined,"string"]: {
    // ...more checks
    indexes = {
      startIndex: Math.max(0, endIndex - (1 + params.pageSize!)),
      endIndex
    };
  }
}

// exit condition
if (indexes === null) {
  res.json({ items: [] });
  return;
}

// from here, work with indexes
```

Undermaken: Thank you, It was a sentiment I had but I want to ignore 🙈 
Your hint it's surely better. In this case I see a clear _pattern-matching_ so I introduced I used [ts-patterns](https://github.com/gvergnaud/ts-pattern)

What do you think? Could it be a good refactoring? 
https://github.com/pagopa/io-dev-api-server/pull/79/commits/7c4d18b8c5a06d8cf58ac21ed4182fcb1e556f04

Undermaken: I did refactoring inverting some condition and remove useless `return undefined`
https://github.com/pagopa/io-dev-api-server/pull/79/commits/a69a770d79bf14087ef804d073740b18519f0184

PR #80, opened 2021-10-12
[IASV-16] Update annullaVoucher API

debiff:
This PR update the `/beneficiario/annullaVoucher ` with respect to the new swagger definition.


src/routers/features/siciliaVola/secured/index.ts
```diff
@@ -122,7 +122,7 @@ addHandler(
         v => v.idVoucher !== maybeVoucherId.value
       );
     }
-    res.sendStatus(200);
+    res.status(200).json({});
```
Undermaken, line 125:
```suggestion
    res.json({});
```
`json` is a shortcut to response with status 200 and a json payload

PR #81, opened 2021-10-13
[IASV-32] update available destination

debiff:
This PR replace the `aeroportiSede` and `aeroportiStato` API with the new `aeroportiAmmessi` API.


src/routers/features/siciliaVola/secured/index.ts
```diff
@@ -81,24 +81,18 @@ addHandler(
...
  */
 addHandler(
   securedSvRouter,
-  "get",
-  addPrefix("/beneficiario/aeroportiStato/:state_id"),
-  (_, res) => res.json(getAereoportiSede(5))
+  "post",
+  addPrefix("/beneficiario/aeroportiAmmessi"),
+  (req, res) => {
+    const { aeroportiAmmessiInputBean } = req.body;
```
Undermaken, line 92:
```suggestion
    const { aeroportiAmmessiInputBean } = req.body;
```
Is it used?

Undermaken (changes requested):

src/routers/features/siciliaVola/secured/index.ts
debiff: Forgotten a commit, fixed in [9ef9962](https://github.com/pagopa/io-dev-api-server/pull/81/commits/9ef99622512d739a6b042d8198633e5506bbbe58)

PR #82, opened 2021-10-13
Remove services metadata as static content

Undermaken:
This PR removes the API to serve service metadata as static content and it adds metadata into the service payload

### DIFF
--- a/src/routers/services_metadata.ts
+++ b/src/routers/services_metadata.ts
@@ -16,6 +16,9 @@ export const servicesMetadataRouter = Router();
 
 const addRoutePrefix = (path: string) => `${staticContentRootPath}${path}`;
 
+/**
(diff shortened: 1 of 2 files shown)


src/routers/services_metadata.ts
```diff
@@ -3,54 +3,31 @@
...
-        [services[idx].service_id.toLowerCase()]: {
-          ...curr,
-          cta: frontMatter1CTASiciliaVola as NonEmptyString
-        }
-      };
-    }
-    return { ...acc, [services[idx].service_id.toLowerCase()]: curr };
-  },
-  {}
-);
```
CrisTofani, line 44: Do we actually need to remove? what if we need to check on a previous version of the App that needs the metadata as they're retrieved nowadays?

CrisTofani (approved):

PR #87, opened 2021-11-03
[IARS-46] Add the support to special services and service category

CrisTofani:
This PR introduces the support services' categories.

`getServices` and related functions to include services like SiciliaVola on the services list construction has been refactored to better include special services on future releases.  

**PLEASE NOTE** the definitions on `io-backend` are still on a WIP state, we need to wait the pagopa/io-backend#841 to be merged in order to merge this PR safely


src/payloads/service.ts
```diff
@@ -115,10 +133,24 @@ export const getServices = (
...
+  const nationalLocalServices: ReadonlyArray<ServicePublic> = [
     ...createService(ServiceScopeEnum.LOCAL, local),
     ...createService(ServiceScopeEnum.NATIONAL, national)
   ];
+
+  const organizationsCount = new Set(
+    nationalLocalServices.map(s => s.organization_fiscal_code)
+  ).size;
+
+  return nationalLocalServices.concat(
```
Undermaken, line 146: I believe this approach could not scale. We should manually increase the counter.
Furthermore `organizationsCount` is only used as size, could be used `nationalLocalServices.length` instead?

My suggestion is based on an array of tuple where the first element is the flag and the second element is the service factory

```typescript
  const specialServices: ReadonlyArray<readonly [
    boolean,
    (orgCount: number) => ServicePublic
  ]> = [
    [ioDevServerConfig.services.includeSiciliaVola, withSiciliaVolaService],
    [ioDevServerConfig.services.includeCgn, withCgnService]
  ];

  // eventually add the special services based on config flag
  return specialServices.reduce((acc, curr) => {
    if (curr[0]) {
      return [...acc, curr[1](acc.length)];
    }
    return acc;
  }, nationalLocalServices);
```
note `specialServices` could be a const outside this function

Undermaken (changes requested):

src/payloads/service.ts
CrisTofani: @Undermaken Good point, this is absolutely a smarter way to include more "special" services on future. I propose another small refactoring on folder structure/file organization. Since I imagine special services may increase on future I think we can reorganize services in this way:
- `payloads/services` <-- this is the main folder
- `payloads/services/index.ts` <-- the current `service.ts`
- `payloads/services/special.ts` <-- here we collect the special services implementation
- `payloads/services/optional.ts` <-- when and if it will be needed we can introduce them here

What do you think?

CrisTofani: 6687b11 includes the proposed refactoring

Undermaken: I believe it is an improvement!
I just done a commit https://github.com/pagopa/io-dev-api-server/pull/87/commits/d299326f5c2eb6c8364f6daf108620960efc9e54 with some refactoring (renaming and moving some functions) and yes, you were right (and I was wrong) when you used the `Set` to compute how many organizations are in the service!
Could you check if it is good?

Anyway should we merge this PR when the new io-backend spec will be officially deployed?

CrisTofani: Yes I think it's better to wait since the api spec may need some rework. 

Talking about the Set, yes we may have more services than organizations 🙈 just forgot about that when I accepted your suggestion

Undermaken: I did another round of refactoring 
https://github.com/pagopa/io-dev-api-server/pull/87/commits/9bbe9a71f08648f1c2139f1aa0626926404f73f9
Double check when you can

CrisTofani: LGTM whenever the API spec is froze we can merge this PR

pietro909:
@Undermaken @CrisTofani can we merge it?

package.json
```diff
@@ -4,8 +4,8 @@
   "description": "A mock server to help io-app development https://io.italia.it/",
   "version": "1.0.0",
   "main": "app.js",
-  "api_backend_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v7.24.0-RELEASE/api_backend.yaml",
-  "api_public_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v7.23.0-RELEASE/api_public.yaml",
+  "api_backend_specs": "https://raw.githubusercontent.com/pagopa/io-backend/special-service-specs/api_backend.yaml",
```
pietro909, line 7: time to point to `7.26.0` 😃

package.json
```diff
@@ -4,8 +4,8 @@
...
+  "api_public_specs": "https://raw.githubusercontent.com/pagopa/io-backend/special-service-specs/api_public.yaml",
```
CrisTofani, line 8:
```suggestion
  "api_backend_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v7.26.0-RELEASE/api_backend.yaml",
  "api_public_specs": "https://raw.githubusercontent.com/pagopa/io-backend/v7.26.0-RELEASE/api_public.yaml",
```

package.json
CrisTofani: I was just changing it ca074d6 😄

CrisTofani: We can now merge this

Undermaken (approved):

PR #88, opened 2021-11-04
[ASZ-109] ZendeskToken api

debiff:
This PR add the `zendeskToken` to the `PublicSession` response.


Undermaken (approved):
This PR LGTM
I believe we should merge it when the backend spec becomes official.
Our dev-server should always acts as the current backend version

What do you think?

debiff:
> This PR LGTM I believe we should merge it when the backend spec becomes official. Our dev-server should always acts as the current backend version
> 
> What do you think?

Yes, it's ok also for me. I wanted to open the PR to speed up the review process so that when the beckend is ready it will be possible to merge

debiff:
@Undermaken Updated with the official specs 💪

Undermaken (approved):

PR #94, opened 2021-11-19
[IA-473] Fix small pagination issue and add config for live mode

pietro909:


src/types/config.ts
```diff
@@ -70,6 +70,12 @@ const HttpResponseCode = t.union([
 ]);
 
 const AllowRandomValue = t.interface({ allowRandomValues: t.boolean });
+const LiveModeMessages = t.interface({
+  // interval between updates in millis
+  interval: t.number,
```
Undermaken, line 75:
```suggestion
  interval: Millisecond,
```
you could use a type that already describes the used unit of measure
`import { Millisecond } from "@pagopa/ts-commons/lib/units";`

Undermaken (changes requested):

src/types/config.ts
pietro909: I'm getting:

> TS2693: 'Millisecond' only refers to a type, but is being used as a value here.

I may be mixing up things here.

Undermaken: this is my bad suggestion!
I thought it is a codec/type
Forget it. Sorry

Undermaken (approved):

PR #107, opened 2021-12-17
[IAMVL-14] Support Legal message category and details

pietro909:


src/payloads/message.ts
```diff
@@ -96,6 +133,11 @@ export const getCategory = (
   const senderService = services.find(
     s => s.service_id === message.sender_service_id
   )!;
+  if ((message as LegalMessageWithContent).legal_message) {
```
Undermaken, line 136:
```suggestion
  if (LegalMessageWithContent.decode(message).isRight()) {
```
pretty the same but using the dedicated codec

src/payloads/message.ts
```diff
@@ -96,6 +133,11 @@ export const getCategory = (
...
+    return {
```
Undermaken, line 137: As you know I like the pattern matching
I tried to convert these if/else statements using it
This is the result

```typescript
  return match(message)
    .when(
      m => LegalMessageWithContent.decode(m).isRight(),
      () => ({
        tag: TagEnumBase.LEGAL_MESSAGE
      })
    )
    .with(
      {
        content: {
          eu_covid_cert: not(__.nullish)
        }
      },
      () => ({
        tag: TagEnumBase.EU_COVID_CERT
      })
    )
    .with(
      {
        content: {
          payment_data: not(__.nullish)
        }
      },
      m => ({
        tag: TagEnumPayment.PAYMENT,
        rptId: getRptID(senderService, m.content.payment_data)
      })
    )
    .otherwise(() => ({
      tag: TagEnumBase.GENERIC
    }));
};
```
Feel free to skip it

Undermaken (changes requested):

src/routers/message.ts
```diff
@@ -424,3 +439,23 @@ addHandler(
...
+addHandler(
+  messageRouter,
+  "get",
+  addApiV1Prefix("/legal-messages/:id"),
+  (req, res) => {
+    // tslint:disable-next-line:no-commented-code
+    // if (configResponse.getLegalMessageResponseCode !== 200) {
+    //   res.sendStatus(configResponse.getLegalMessageResponseCode);
+    //   return;
+    // }
```
Undermaken, line 452: could it be removed?

src/routers/message.ts
```diff
@@ -424,3 +439,23 @@ addHandler(
...
+
+    // retrieve the messageIndex from id
+    const message = messagesWithContent.find(item => item.id === req.params.id);
+    if (message === undefined) {
+      res.json(getProblemJson(404, "message not found"));
```
Undermaken, line 457:
```suggestion
      res.json(getProblemJson(404, "message not found"));
      return;
```

src/routers/message.ts
```diff
@@ -424,3 +439,23 @@ addHandler(
...
+    }
+    res.json(message);
```
Undermaken, line 459: Should we also check that the request message is a legal message?
```suggestion
    if (LegalMessageWithContent.decode(message).isLeft()) {
      res.json(getProblemJson(400, "bad parameter"));
      return;
    }
    res.json(message);
```

Undermaken (changes requested):

src/payloads/message.ts
```diff
@@ -48,6 +52,39 @@ export const withDueDate = (
...
+        sender: "dear.leader@yourworld.hell" as NonEmptyString,
+        recipients: "slave@yourworld.hell" as NonEmptyString,
+        replies: "down.the.john@theshitIgive.hell" as NonEmptyString,
+        object: "I told you already: you're fired" as NonEmptyString
+      },
+      data: {
+        sender_provider: "apocalypse knights" as NonEmptyString,
+        timestamp: (new Date().toISOString() as unknown) as UTCISODateFromString,
+        envelope_id: "abcde" as NonEmptyString,
+        msg_id: "fghi" as NonEmptyString,
```
Undermaken, line 77:
```suggestion
        msg_id: `mvl_${message.id}` as NonEmptyString,
```
Maybe we should not use a fixed id for all MVL messages.

Undermaken (changes requested):

src/payloads/message.ts
pietro909: I love pattern matching too, though it is not supported by the language therefore I'd rather use something more idiomatic like truthy checks and early returns.

src/routers/message.ts
pietro909: It's a good observation, though I don't know how the API would respond in this case.
Maybe @fabriziopapi can answer this question.

src/routers/message.ts
pietro909: I left it there for [future use](https://github.com/pagopa/io-dev-api-server/pull/107/files#diff-c4cdb7cf0e69ecfc696d0ba80b8411dd4297658be4a57dad5f0954d7a9ae6d74R448). Now is working 🎉

src/routers/message.ts
Undermaken: We already know what this API returns in case of `200`
https://github.com/pagopa/io-backend/blob/df2c4bb5a62df3ce2f9254c04ca56bae18032032/api_backend.yaml#L280
I believe we should ensure that from the given id we are actually returning a legal message detail

Undermaken (changes requested):

src/routers/message.ts
pietro909: I think we are overthinking it: if the point of the dev-server is to mimick the actual API, we must know how it would behave if a client asked for an ID which is a message but not a _legal_ one.
Otherwise, 404 is totally fine in my opinion.

pietro909: Updated in https://github.com/pagopa/io-dev-api-server/pull/107/commits/4373e149d34a0d2d4b754b8b0e1e370ea9b25777 as per our chat. Now returning a meaningful message which will help us during development.

Undermaken:
This PR will be included in https://github.com/pagopa/io-dev-api-server/pull/108

PR #113, opened 2021-12-30
[IA-428] Display local services with organization and service names

Undermaken:
## description
This PR improves the visualization of local service web page (embedded in IO app)

| before  | now |
| ------------- | ------------- |
before | Privacy

### DIFF
--- a/assets/html/services_web_view.html
+++ b/assets/html/services_web_view.html
@@ -23,9 +23,9 @@
           groupByOrganizationName[s.organization_name] = [...groupByOrganizationName[s.organization_name],s];
         });
         Object.keys(groupByOrganizationName).forEach(orgKey => {
-          container.append(`<div><h1 style="font-size: 40px;color: black;">${orgKey}</h1></div>`);
+          container.append(`<div><h1 style="font-size: 22px;color: black;">${orgKey}</h1></div>`);
           groupByOrganizationName[orgKey].forEach(s => {
-            container.append(`<div><h2 style="font-size: 30px;color: blue; cursor: pointer;" onclick="sendMessagesToRN('${s.service_id}')">${s.service_name}</h2></div>`);
+            container.append(`<div><h2 style="font-size: 18px;color: blue; cursor: pointer;" onclick="sendMessagesToRN('${s.service_id}')">${s.service_name}</h2></div>`);
           });
         });
         }
@@ -41,7 +41,7 @@
   </script>
 </head>
 <body style="padding:10px;">
-<div style="font-size: 60px;">DEV-SERVER local services</div>
+<div style="font-size: 30px;">DEV-SERVER local services</div>
 <div id="container"></div>
 
 </body>
--- a/assets/html/services_web_view.html
+++ b/assets/html/services_web_view.html
@@ -22,12 +22,14 @@
           }
           groupByOrganizationName[s.organization_name] = [...groupByOrganizationName[s.organization_name],s];
         });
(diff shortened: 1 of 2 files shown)


assets/html/services_web_view.html
```diff
@@ -14,9 +14,19 @@
 
     function loadServices(){
       const container = $("#container");
-      $.getJSON("/api/v1/services", function(services) {
-        services.items.filter(p => p.scope === "LOCAL").forEach(s => {
-          container.append(`<div><h1 style="font-size: 40px;color: blue; cursor: pointer;" onclick="sendMessagesToRN('${s.service_id}')">${s.service_id}</h1></div>`);
+      $.getJSON("/services_web_view/local_services", function(services) {
```
pietro909, line 17: Wouldn't it be better to keep the filtering here? To remove one line of code we add [a new unofficial handler](https://github.com/pagopa/io-dev-api-server/pull/113/files#diff-8dd6b29c166231bd5f90ad361cf9143c34c51fccbf08a719e4835a92e43ede91R130) and my understanding is that it goes against the idea of having a dev-server as close as possible to the real life backend.

assets/html/services_web_view.html
```diff
@@ -14,9 +14,19 @@
...
+          if(!groupByOrganizationName[s.organization_name]){
+            groupByOrganizationName[s.organization_name] = [];
+          }
+          groupByOrganizationName[s.organization_name] = [...groupByOrganizationName[s.organization_name],s];
+        });
+        Object.keys(groupByOrganizationName).forEach(orgKey => {
+          container.append(`<div><h1 style="font-size: 22px;color: black;">${orgKey}</h1></div>`);
+          groupByOrganizationName[orgKey].forEach(s => {
+            container.append(`<div><h2 style="font-size: 18px;color: blue; cursor: pointer;" onclick="sendMessagesToRN('${s.service_id}')">${s.service_name}</h2></div>`);
+          });
```
pietro909, line 29: Not a big deal here, but in FE dev the DOM access is normally reduced as much as possible being very slow compared to JS execution.
A common pattern is to concatenate the string in the loop(s) and update the DOM only at the very end.

assets/html/services_web_view.html
Undermaken: The local services in the app are loaded in the webview, so there is no interaction with the backend.
That [web page](https://io.italia.it/app-content/enti-servizi.html) is filled with [a static JSON](https://io.italia.it/assets/json/enti-list-webview.json) (built every night) that expose only the local services

Actually the official backend API returns to the app only the NATIONAL services (first tab in the service app section) as the dev-server-does.

If we have to follow the official backend API this page has to do (X+1) requests to build the list

- 1 to get all services ID (`/services`) _<- but actually it return only NATIONAL_
- X for each services in the previous list to get the details (`/services/XServiceID`)

I believe this scenario is almost close to the real one where the webpage has no interaction with the IO backend

assets/html/services_web_view.html
Undermaken: Thank you, I didn't know.
Changed https://github.com/pagopa/io-dev-api-server/pull/113/commits/01c0bd51fef4edfa58c882e70c24b77c083a3373

assets/html/services_web_view.html
pietro909: Fine, thank you for the explanation!

pietro909 (approved):

PR #114, opened 2022-01-04
[IA-614] Handle paypal remote config

Undermaken:
# ⚠️ this PR depends on https://github.com/pagopa/io-services-metadata/pull/536 
This PR handles the remote config for paypal

### DIFF
--- a/src/payloads/backend.ts
+++ b/src/payloads/backend.ts
@@ -208,6 +208,9 @@ export const backendStatus: BackendStatus = {
     cgn_merchants_v2: false,
     assistanceTool: {
       tool: ioDevServerConfig.assistanceTools.tool
+    },
+    paypal: {
+      enabled: ioDevServerConfig.wallet.paypalEnabled
     }
   }
 };
--- a/src/payloads/backend.ts
+++ b/src/payloads/backend.ts
@@ -201,7 +201,7 @@ export const backendStatus: BackendStatus = {
   config: {
     bpd: {
       enroll_bpd_after_add_payment_method: false,
-      program_active: true
+      program_active: false
     },
     bpd_ranking: true,
     bpd_ranking_v2: true,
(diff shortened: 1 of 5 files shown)


src/payloads/backend.ts
```diff
@@ -201,13 +201,16 @@ export const backendStatus: BackendStatus = {
...
+      program_active: false
     },
     bpd_ranking: true,
     bpd_ranking_v2: true,
     cgn_merchants_v2: false,
     assistanceTool: {
       tool: ioDevServerConfig.assistanceTools.tool
+    },
+    paypal: {
+      enabled: ioDevServerConfig.wallet.paypalEnabled
```
fabriziofff, line 213: In this case the file `src/payloads/backend.ts` is already a configuration file that represents the returned response.
I think could be a little confusing take some values (`assistanceTool.tool` and `paypal.enabled`) from another configuration file with a different path (`config.paypal.enabled` -> `wallet.paypalEnabled`) and manage the increase complexity of a similar mapping.
Since in case like this we don't need a value from the config to change the behaviour of the dev server but only to directly configure the returned json, what you think about keep considering the file `src/payloads/backend.ts` as a configuration file and using already the required values? 
In the future we can think to have differents `backend.json` and in the dev server config can be configured the path of the desired `backend.json` to use as response, like
```typescript
{
   ...
   backendResponse: "./assets/backend/defaultResponse.json"
   ...
}
```

fabriziofff (changes requested):

src/payloads/backend.ts
Undermaken: Good point.
I believe I want a little off-road due to make the dev-server iper-configurable.
What about this commit 908fbb56b28396d17c0bbf364cfadade07b8febe?

fabriziofff (approved):

PR #115, opened 2022-01-07
Implements handler for special services activation

CrisTofani:
This PR aims to properly handle the special services activation status. 
CGN activation now effects the related service activation

### DIFF
--- a/src/payloads/services/index.ts
+++ b/src/payloads/services/index.ts
@@ -1,5 +1,6 @@
 import * as faker from "faker/locale/it";
 import { range } from "fp-ts/lib/Array";
+import { fromNullable } from "fp-ts/lib/Option";
 import {
   NonEmptyString,
   OrganizationFiscalCode
@@ -10,6 +11,9 @@ import { ServiceId } from "../../../generated/definitions/backend/ServiceId";
 import { ServicePreference } from "../../../generated/definitions/backend/ServicePreference";
 import { ServicePublic } from "../../../generated/definitions/backend/ServicePublic";
 import { ServiceScopeEnum } from "../../../generated/definitions/backend/ServiceScope";
+import { SpecialServiceMetadata } from "../../../generated/definitions/backend/SpecialServiceMetadata";
+import { ioDevServerConfig } from "../../config";
+import { isCgnActivated } from "../../routers/features/cgn";
 import { getService, getServiceMetadata } from "../../utils/service";
 import { validatePayload } from "../../utils/validator";
 import { frontMatter2CTA2 } from "../../utils/variables";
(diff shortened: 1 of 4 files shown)


src/payloads/services/index.ts
```diff
@@ -74,11 +78,39 @@ export const getServicesTuple = (
...
+>([["cgn", isCgnActivated]]);
+
 export const getServicesPreferences = (
   services: ReadonlyArray<ServicePublic>
 ) =>
   new Map<ServiceId, ServicePreference>(
     services.map(s => {
+      const metadata = s.service_metadata;
+      if (metadata && SpecialServiceMetadata.is(metadata)) {
+        const specialServiceInbox = fromNullable(metadata.custom_special_flow)
```
pietro909, line 93: It seems like `specialServiceInbox` is a boolean: maybe `hasSpecialServiceInbox` would better convey this semantics?

src/payloads/services/index.ts
```diff
@@ -74,11 +78,41 @@ export const getServicesTuple = (
...
+              h()
+            )
+          )
+          .getOrElse(false);
+        return [
+          s.service_id,
+          {
+            is_inbox_enabled: hasSpecialServiceInbox,
+            is_email_enabled: hasSpecialServiceInbox
+              ? faker.datatype.boolean()
```
Undermaken, line 107:
```suggestion
              ? getRandomValue(false, faker.datatype.boolean(), "services")
```
we are trying to dominate the randomness in the dev-server with the configuration
A dev could go crazy if a value change on every request :dizzy_face:

src/routers/features/cgn/index.ts
```diff
@@ -91,6 +98,22 @@ addHandler(cgnRouter, "get", addPrefix("/activation"), (_, res) =>
...
       };
+
+      const currentPreference = servicesPreferences.get(
+        cgnServiceId as ServiceId
+      );
+
+      const increasedSettingsVersion = currentPreference
+        ? (((currentPreference.settings_version as number) +
+            1) as ServicePreference["settings_version"])
+        : (0 as ServicePreference["settings_version"]);
```
Undermaken, line 109:
```suggestion
      const increasedSettingsVersion = ((currentPreference?.settings_version ??
        -1) + 1) as NonNegativeInteger;
```
the same, shorter but less readable. On your flavor

src/routers/features/cgn/index.ts
```diff
@@ -91,6 +98,22 @@ addHandler(cgnRouter, "get", addPrefix("/activation"), (_, res) =>
...
+      servicesPreferences.set(cgnServiceId as ServiceId, {
+        is_inbox_enabled: true,
+        is_email_enabled: faker.datatype.boolean(),
```
Undermaken, line 112:
```suggestion
        is_email_enabled:currentPreference?.is_email_enabled ??
          getRandomValue(false, faker.datatype.boolean(), "services"),
```
the same of the other part: we should avoid the dev madness with a value that could change on every API invocation

Undermaken (changes requested):

src/payloads/services/index.ts
```diff
@@ -74,11 +80,47 @@ export const getServicesTuple = (
...
+
+        return [
+          s.service_id,
+          {
+            is_inbox_enabled: hasSpecialServiceInbox,
+            is_email_enabled: hasSpecialServiceInbox
+              ? getRandomValue(false, faker.datatype.boolean(), "services")
+              : false,
+            is_webhook_enabled: hasSpecialServiceInbox
+              ? faker.datatype.boolean()
```
Undermaken, line 118:
```suggestion
              ? getRandomValue(false, faker.datatype.boolean(), "services")
```

Undermaken (changes requested):

Undermaken (approved):

PR #132, opened 2022-01-31
[IAI-120] Add versionInfo endpoint

fabriziofff:
This pr adds the new `status/versionInfo.json` endpoints. See https://github.com/pagopa/io-services-metadata/pull/554 for more information.


src/routers/services_metadata.ts
```diff
@@ -45,6 +45,14 @@ addHandler(
   (_, res) => res.json(backendStatus)
 );
 
+// Metadata related to the app version
+addHandler(
+  servicesMetadataRouter,
+  "get",
+  addRoutePrefix("/status/versionInfo.json"),
+  (_, res) => sendFile("assets/status/versionInfo.json", res)
```
Undermaken, line 53: If we add the new definition of services-metadata in `packages.json`
`1.0.4 -> 1.0.5`
`https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.5/definitions.yml`

we could improve the payload check to avoid wild payloads sent to the clients

```suggestion
(_, res) => {
    pipe(
      readFileAsJSON("assets/status/versionInfo.json"),
      VersionInfo.decode,
      E.fold(
        errors => res.status(500).send(readableReport(errors)),
        v => res.json(v)
      )
    );
  }
```

Undermaken (changes requested):

src/routers/services_metadata.ts
fabriziofff: Yes, much better and with `fp-ts 2.x` 😍 
Added a new function `readFileAndDecode` in 43eb149

Undermaken (approved):

PR #137, opened 2022-02-03
[IAC-139] Update BPS citizenV2 GET/PUT API

Undermaken:
This PR depends on https://github.com/pagopa/io-services-metadata/pull/561


src/routers/features/bdp/__tests__/index.test.ts
```diff
@@ -27,4 +37,20 @@ describe("citizen V2 API", () => {
       expect(cr.value.enabled === true).toBeTruthy();
     }
   });
+
+  it("Should return a 200, CitizenResource (V2) with optInStatus update", async () => {
```
debiff, line 41: The test looks good, but if I change the order of the tests, or if I execute it in isolation, it fails.

Maybe we can group the tests that need the same status under a `describe` and use the `beforeEach` function to lead to the right status?

debiff (changes requested):

src/payloads/backend.ts
```diff
@@ -210,7 +210,7 @@ export const backendStatus: BackendStatus = {
   config: {
     bpd: {
       enroll_bpd_after_add_payment_method: false,
-      program_active: false,
+      program_active: true,
```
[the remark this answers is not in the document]
Undermaken: 🦅 
good catch

src/routers/features/bdp/__tests__/index.test.ts
Undermaken: You hit my laziness, thank you
refactored here 
https://github.com/pagopa/io-dev-api-server/pull/137/commits/50a8bdd0c77cbf4a2a514d38e355d9d466883851
Schermata 2022-02-08 alle 19 20 47

debiff (approved):

PR #143, opened 2022-03-01
[IA-678] Support archived and read status server side

pietro909:
## Short description
Introduce a new endpoint to mark messages as read or archived. Persist the information on the server.
API specs available at https://raw.githubusercontent.com/pagopa/io-backend/83b522a8b83d551c773560ea579c76e78a2ef288/api_backend.yaml

* implement the [in-memory persistence layer](src/persistence/messages.ts)
* use [updated API specs](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R7)
* add [dev script](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R33) for testing (same as IO App)
* add [new config parameter](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R85) to return archived messages
* move mock data generation to [its own script](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-c22895b0bd25878cee9a78c4f6059ad8e9f70d4fa89fd8008c3890f1c077979bR318) accepting custom config
* add [new PUT endpoint](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-c4cdb7cf0e69ecfc696d0ba80b8411dd4297658be4a57dad5f0954d7a9ae6d74R167) with [tests](https://github.com/pagopa/io-dev-api-server/pull/143/files#diff-1921bcbd0302c2445f41a744a63f3450ad404cc85b48e9474ab5e1f520d020e2R202) to upsert message statuses

## How to test

The updates to the model and the persistence layer can be tested using [this branch of IO App](https://github.com/pagopa/io-app/pull/3805).
The new endpoint must be tested manually until [IA-728](https://pagopa.atlassian.net/browse/IA-728) is delivered.


src/persistence/messages.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+// tslint:disable-next-line: readonly-array no-let
+function persist(messages: MessageFromBE[]): void {
+  inboxMessages = inboxMessages
+    .concat(messages.map(m => ({ ...m, is_read: false, is_archived: false })))
+    .sort((a, b) => (a.id < b.id ? 1 : -1));
+}
+
+function findAllInbox(): ReadonlyArray<MessageOnDB> {
+  return inboxMessages as ReadonlyArray<MessageOnDB>;
```
Undermaken, line 61:
```suggestion
  return inboxMessages;
```
maybe it is no need since it is declared in the return type

src/persistence/messages.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function findAllArchived(): ReadonlyArray<MessageOnDB> {
+  return archivedMessages as ReadonlyArray<MessageOnDB>;
```
Undermaken, line 65:
```suggestion
  return archivedMessages;
```
same of above

src/persistence/messages.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Find one message given its ID, whether it is in the inbox or archived.
+ */
+function findOneById(id: string): MessageOnDB | null {
+  return (
+    inboxMessages.find(message => message.id === id) ||
+    archivedMessages.find(message => message.id === id) ||
+    null
+  );
```
Undermaken, line 76:
```suggestion
  return (
    [...inboxMessages, ...archivedMessages].find(
      message => message.id === id
    ) ?? null
  );
```
pretty the same, choose to your taste
PS: if you return undefined (as `find` already does) we can avoid the _nullish coalescing_

src/populate-persistence.ts
```diff
@@ -0,0 +1,342 @@
...
+  if (services.length === 0) {
+    throw new Error(
+      "to create messages, at least one sender service must exist!"
+    );
+  }
+  return faker.random.arrayElement(services).service_id;
+};
+
+const getNewMessage = (
+  customConfig: IoDevServerConfig,
```
Undermaken, line 43: if you want you can avoid this parameter since the config is available as object from everywhere 
`ioDevServerConfig`

src/populate-persistence.ts
```diff
@@ -0,0 +1,342 @@
...
+} from "./utils/variables";
+
+const getRandomServiceId = (): string => {
+  if (services.length === 0) {
+    throw new Error(
+      "to create messages, at least one sender service must exist!"
+    );
+  }
+  return faker.random.arrayElement(services).service_id;
+};
```
Undermaken, line 40:
```suggestion
const getServiceId = (): string => {
  if (services.length === 0) {
    throw new Error(
      "to create messages, at least one sender service must exist!"
    );
  }
  return getRandomValue(
    services[0].service_id,
    faker.random.arrayElement(services).service_id,
    "messages"
  );
};
```
to make the global config more efficient, could be a good idea to return a random `serviceId` only when the randomness is enabled for this specific section (`messages`)?

src/populate-persistence.ts
```diff
@@ -0,0 +1,342 @@
...
+      medicalPrescription
+    );
+
+  const barcodeReceipt = fs
+    .readFileSync("assets/messages/barcodeReceipt.svg")
+    .toString("base64");
+
+  /* medical */
+  range(1, customConfig.messages.medicalCount).forEach(count => {
+    output.push(medicalMessage(count));
```
Undermaken, line 148:
```suggestion
```
double push

Undermaken (changes requested):

src/persistence/messages.ts
pietro909: I'd keep the `null` because it reflects how (some) SQL databases work.
No real opinions on the rest, just tell me if you want me to change it.

src/populate-persistence.ts
pietro909: No opinions from my side: is a testing tool and I see no value in performance tuning.
Your call 😃

Undermaken: What I meant with `Efficient` is not about performance but to make the dev-server deterministic (or with a certain degree of randomness) by configuration. 

As you can see each sections of the configuration (messages, profile, services etc..) has an optional flag `allowRandomValue` . This is useful in dev-environment (or in automatic e2e tests) when you want to load always the same data and avoid randomness.

Undermaken (changes requested):
To avoid dangerous type casting (`any` and `unknown`) I tried, in a dedicated branch, to refactor some types and it seems to work
4b83e1677b673fca331880c67912ba1ef75bd41b
What do you think?

src/populate-persistence.ts
pietro909: Got it. I was misled by _efficient_.

pietro909:
@Undermaken if it compiles and feels safer, just commit it 😄

Undermaken (approved):

PR #148, opened 2022-03-10
The payment activation failure config is ignored

Undermaken:
## Short description
This PR introduces the handling of the activation failure by config


src/routers/payment.ts
```diff
@@ -159,6 +173,7 @@ export const handlePaymentPostAndRedirect = (
   outcomeValue: number = 0,
   title: string = "Pay web page"
 ) => {
+  console.log(req.body);
```
pietro909, line 176: 🚨 **issue**: the console log is not supposed to be here, right?

Undermaken: Yes it is.
Forgot after some test.
Thank you 676b8f255e14db6287d7ebc91466547976c63ecb

pietro909 (approved):

PR #150, opened 2022-03-14
[IA-722] Improved donations page accessibility

dgopsq:
## Short description
This PR is going to improve the accessibility in the donations page.

**N.B.** There are hard-coded values like `Pulsante di scelta` because using the native elements inside the web-view would cause various words to not be translated in Italian.

https://user-images.githubusercontent.com/5963683/158809171-64a941d8-8511-43a2-ae49-46837d9ebd8c.mov

- Removed the `input + label` hack in favour of pure JavaScript, since it would cause accessibility issues
- Added more `aria` attributes
- Added a custom label to the charity radio button and the donation amounts
- Added automatic focus on various elements in order to have a better navigation flow

## How to test
Run the donation page using the device's Voice Over, the experience should match the one described in the ticket.

### DIFF
--- a/assets/html/donate.html
+++ b/assets/html/donate.html
@@ -493,7 +493,7 @@
       replaceChildren(list, [selectedCharity]);
       selectedCharity.classList.add("selected");
 
-      document.getElementById("tiers-container").querySelector("li.tier .tier-item").focus()
+      document.getElementById("how-much-title").focus()
     } catch (e) {
       reportError(e);
     }
--- a/assets/html/donate.html
+++ b/assets/html/donate.html
@@ -412,7 +412,7 @@
 
   function parseEntities(entities) {
     return entities.map(entity => {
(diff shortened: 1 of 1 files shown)


Undermaken (changes requested):
comparison

@dgopsq sorry I missed to insert a message within the image
Referring to the above comparison: are these a regressions?

- font changed
- a blue rectangle around "Quanto vuoi donare?"

dgopsq:
> comparison
> 
> @dgopsq sorry I missed to insert a message within the image Referring to the above comparison: are these a regressions?
> 
> * font changed
> * a blue rectangle around "Quanto vuoi donare?"

@Undermaken You were right about the fonts, I fixed the error in the latest commits. I couldn't reproduce the outline though, you were getting it from inside the IO app webview? I added the generic CSS rule to disable that behaviour anyway so it could be already solved!

Undermaken (approved):

Undermaken:
Now it looks fine!
@dgopsq Could you create a task (and execute if you want) to sync these changes in `io-services-metadata`?

PR #169, opened 2022-05-05
[IOACGN-72] Supports the information about new discount on categories, merchants list and merchant detail

CrisTofani:
## Short description
This PR adds the support to the information of new discounts published by merchants on Categories list, merchant item on search and merchant detail

### DIFF
--- a/src/routers/features/cgn/merchants.ts
+++ b/src/routers/features/cgn/merchants.ts
@@ -129,6 +129,15 @@ addHandler(
     ].flatMap(item => item.productCategories);
     const categoriesSet = new Set(categories);
 
+    if (req.query.count_new_discounts) {
+      res.json({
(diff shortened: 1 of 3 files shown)


src/routers/features/cgn/merchants.ts
```diff
@@ -129,6 +129,15 @@ addHandler(
     ].flatMap(item => item.productCategories);
     const categoriesSet = new Set(categories);
 
+    if (req.query.count_new_discounts) {
```
Undermaken, line 132:
```suggestion
    if (BooleanFromString.decode(req.query.count_new_discounts).getOrElse(false)) {
```
I guess `count_new_discounts` will be always a string, since it is the query string
Does it make sense?

Undermaken (changes requested):

Undermaken (approved):

PR #185, opened 2022-07-18
[SFEQS-912] Add Firma con IO feature flag

hevelius:
## Short description
This PR adds the `fci` feature flag reproducing the prod config

- Added the `fci` ff set to false


francescopersico:
@hevelius Can you review this PR, code changes seems strage. Probably something got mess with merges.

hevelius:
Since the remote feat flag is already present on services-metadata, it has been added to the master to align the payload. I believe this PR is no longer needed. What do you think about this @francescopersico ?

hevelius:
It should be synced with this aa00920ce3edff54f23df1f6be7e287165db5553

PR #188, opened 2022-10-13
[IAI-249] Update README instructions to include API definitions generation

Vangaorth:
## Short description
This PR adds a missing instruction point to the Setup section in the README file. Such point is the API definition generation, after the `yarn install` command.

- Added the API definitions generation instruction and command on the SETUP section of the README file

## How to test
Follow the Setup process after repository cloning

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -23,6 +23,8 @@ The node version used in this project is stored in [.node-version](.node-version
 1. clone this repository
 1. install all packages needed:<br/>
    `yarn install`
+1. generate API definitions:
+   `yarn generate:all`
 1. you are ready! start the server:<br/>
    `yarn start`
 
--- a/README.MD
+++ b/README.MD
@@ -23,7 +23,7 @@ The node version used in this project is stored in [.node-version](.node-version
 1. clone this repository
 1. install all packages needed:<br/>
    `yarn install`
-1. generate API definitions:
+1. generate API definitions:<br/>
    `yarn generate:all`
 1. you are ready! start the server:<br/>
    `yarn start`


README.MD
```diff
@@ -23,6 +23,8 @@ The node version used in this project is stored in [.node-version](.node-version
 1. clone this repository
 1. install all packages needed:<br/>
    `yarn install`
+1. generate API definitions:
```
CrisTofani, line 26:
```suggestion
1. generate API definitions:<br/>
```
Just adding a break to make it look similar to the other points 🙂

CrisTofani (changes requested):

README.MD
Vangaorth: Is it possibile that the <br/>-s  are misplaced here? They seem to be ignored when displayed and other lists do not have them (e.g, the Docker Setup below). If this is the case, I would proceed by removing them all

CrisTofani: They're actually not ignored 😄 
That's how it looks like on Github the preview
image

without the `<br/>` block it is represented on the single line, anyway we could rewrite this as an entire code block 🤔 

For example: 

```
# 1. clone this repository
$ git clone git@github.com:pagopa/io-dev-api-server.git
# 2. install all packages needed
$ yarn install
# 3. generate API definitions: 
$ yarn generate:all
# 4. you are ready! start the server:
$ yarn start
```

What do you think?

Vangaorth: While the suggested code-block approach sounds interesting, I would prefer to add the missing `<br/>` since:
- the not-ignored `<br/>` was a mistaken assumption of mine 
- consistency would require us to rewrite any list using the code-block
- existing lists are auto-enumerated, while the code-block requires further investigation to do so

So, if it is good for you, I will proceed with the missing `<br/>`

CrisTofani: Lgtm to add the break 😉

francescopersico (approved):

CrisTofani (approved):

PR #201, opened 2023-01-13
[IOAPPCOM-45] Remote attachments for generic messages

Vangaorth:
## Short description
This PR enhances the dev-server, allowing to generate messages that have third-party attachments.

- `src/config` has a new property `withRemoteAttachments` that generates _n-messages_ with remote attachments (default 0)
- `src/populate-persistence` starts the generation of messages with remote attachments and adds them to the in-memory DB
- `src/payload/message` has the code that generates the message with remote-data and the list of attachments
- `src/router/message` handles single attachment retrieval and sending to the client, distinguishing between PN attachments and generalised remote attachments. Some requests in this file have also been fixed since they were sending a 200 successful HTTP response when a failure was expected
- `src/types/config` adds the specific for `withRemoteAttachments` property
- `assets/messages/remote/attachments` folder contains small sample _.pdf_ attachments that are returned to the client

## How to test
Start the dev server and test the following requests (make sure to test `withRemoteAttachments` with at least a value of 0, a value between 1 and 10, 10, and a value greater than 10):
- `messages`: to retrieve the generic list of messages (`withRemoteAttachments` messages should have a `GENERIC` category's tag)
- `messages/:id`: should have `content.third_party_data.has_attachments` set to `true` for `withRemoteAttachments` messages. For standard messages, `third_party_data` should not appear.
- `third-party-messages/:id` should send the `third_party_message` details, with the `third_party_message.attachments` containing the list of attachments, for `withRemoteAttachments` messages. It should return a `404` error for standard messages
- `third-party-messages/:id/attachments/:attachmentId`: should send the attachment content or return a:
---`404` status code if there is a mismatch between `:id` and `:attachmentId`
---`415` status code if the attachment file is not a PDF 
---`500` status code if the attachment file cannot be found (code is not the best but it is aligned with how the real backend behaves)

Repeat the previous requests for `PN` messages, in order to check that that feature still works as expected (to enable `PN`, set both `includePn` and `pnCount` in the config file)


src/payloads/message.ts
```diff
@@ -1,5 +1,6 @@
 import { FiscalCode, NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import * as faker from "faker";
+import _ from "lodash";
```
hevelius, line 3: If slice is the only method used what do you think about import just required function?
`import { slice } from "lodash";`

src/routers/message.ts
```diff
@@ -5,7 +5,7 @@ import { pipe } from "fp-ts/lib/pipeable";
 import _ from "lodash";
 import { __, match, not } from "ts-pattern";
 import { LegalMessageWithContent } from "../../generated/definitions/backend/LegalMessageWithContent";
-import { MessageContentWithAttachments } from "../../generated/definitions/backend/MessageContentWithAttachments";
+import { TagEnum } from "../../generated/definitions/backend/MessageCategoryPN";
```
hevelius, line 8: What do you think about using something like `CategoryTagEnum` or `PNCategoryTagEnum?`

src/payloads/message.ts
Vangaorth: Done it

src/routers/message.ts
Vangaorth: I do not understand this. Code is later used to compare the tag property of a message's category (to build an asset relative file system path):

```
const messageCategory = getCategory(message);
const categoryTag = messageCategory?.tag;
const attachmentFolderName = categoryTag === TagEnum.PN ? "pn" : "remote";
```

`messageCategory` is of type `MessageCategory`, which has a property `tag` of type `TagEnum`.

What changes are you suggesting?

hevelius: I suggested using named imports to make the name more explicit. So TagEnum could be imported with
`import { TagEnum as PNCategoryTagEnum } from "../../generated/definitions/backend/MessageCategoryPN";`

Vangaorth: Renamed!

hevelius (approved):
LGTM

PR #205, opened 2023-01-25
[IOAPPCIT_35,36] Update dev server flow to support autocomplete login

SabPicariello:
## Short description
Added a post login redirect (both on success and error state) to ensure the proper functioning of the component that enables password autocomplete.
- src/routers/public.ts: added two new redirects that change only the url scheme.

## How to test
Login or test error and verify if a redirect to a url with "ioit" urlScheme occurs.


CrisTofani:
@SabPicariello I changed the title of the PR to properly represent the ticket number whose referring 😄

PR #210, opened 2023-02-10
[IOAPPFD0-30] - Add missing check to the persistence of messages with remote attachments

shadowsheep1:
## Short description
This PR fixes the persistence of messages with remote attachments according with the server configuration.

- src/populate-persistence.ts: add missing check.

## How to test
Run io-app against the dev server configuring 0 messages with remote attachments and see that no message of this kind is present. Add some messages with remote attachments in configuration and check that those messages will be shown.

### DIFF
--- a/src/populate-persistence.ts
+++ b/src/populate-persistence.ts
@@ -398,17 +398,18 @@ const createMessages = (
       );
     });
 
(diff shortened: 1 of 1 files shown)


Vangaorth (approved):
LGTM

PR #223, opened 2023-03-31
[IODPAY-204] Add IDPay IBAN endpoints

mastro993:
## Short description
This PR adds the routes for IDPay and introduces the IBAN endpoints.

IDPay endpoints will be accessible with `http://127.0.0.1:3000/idpay/* `


src/routers/features/idpay/iban.ts
```diff
@@ -0,0 +1,27 @@
+import { pipe } from "fp-ts/lib/function";
+import * as O from "fp-ts/lib/Option";
+import { getIbanResponse } from "../../../payloads/features/idpay/iban/get-iban";
```
hevelius, line 3: Maybe this could be deleted. Only the iban list could be used. What do you think?

mastro993: Yeah, that's right! Will refactor in the next commit 😃

mastro993: Fixed in https://github.com/pagopa/io-dev-api-server/pull/223/commits/d60ac4bfa48882aad5f223b5e9e08a6fddcad2e8

hevelius (approved):
LGTM

PR #225, opened 2023-04-01
chore(Firma con IO): [SFEQS-1579] Update FCI specs

hevelius:
## Short description
This PR adds the new FCI spec that introduced the IssuerEnvironment status value used to witch between TEST and PROD environment QTSP side.

- Update FCI spec
- Update mocked response

## How to test
The tests should be resolved correctly.


mastro993 (approved):
LGTM

PR #227, opened 2023-04-05
chore(Firma con IO): [SFEQS-1590] Add signature requests

hevelius:
## Short description
This PR adds signature-requests endpoint and related payload.

- Add signature-requests payload
- Update test and router
- Add an utils function to get random value from an enum type

## How to test
Tests should be resolved correctly


src/payloads/utils/random.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Returns a random value from an enum.
+ */
+export const getRandomEnumValue = <T extends object>(
+  enumObj: T
+): T[keyof T] => {
+  const enumValues = Object.values(enumObj);
+  const index = Math.floor(Math.random() * enumValues.length);
+
+  return enumValues[index];
+};
```
hevelius, line 11: @mastro993 What do you think about this function to get an enum value randomly?

mastro993 (approved):
LGTM.
All tests resolved correctly

src/payloads/utils/random.ts
mastro993: Yes I like it. It might come in handy in several cases!

PR #243, opened 2023-05-04
chore(Firma con IO): [SFEQS-1640] Add errors response code

hevelius:
## Short description
This PR adds a config item to manage error response code. Generally all response codes not equal to 200 are managed as generic errors by the app. Only 404 is used to catch the case related to an invalid signatureRequestId.

- Update router for FCI
- Update test
- Update config
- Add errorWithStatusCode function

## How to test
Tests should solve correctly. Also it should possibile using the new config with getFciResponseCode !== 200 to get an error screen on the app when starts a signing flow.


Vangaorth (changes requested):

Vangaorth (approved):

Vangaorth:
LGTM!

PR #244, opened 2023-05-04
Added authentication redirect with iologin urlScheme

SabPicariello:
## Short description
This PR edits the redirect urlScheme at the end of login phase, to allow the authentication with native component.

- src/payloads/login.ts: added the new scheme;
- src/routers/public.ts: read `Short description`

## How to test
Try to login with devServer and check if the final redirect starts with `iologin://`

### DIFF
--- a/src/payloads/login.ts
+++ b/src/payloads/login.ts
@@ -3,3 +3,4 @@ export const loginLolliPopRedirect: string = "/idp-login";
 const redirectUrl: string = "/profile.html?token=";
 export const errorRedirectUrl: string = "/error.html?errorCode=";
 export const loginWithToken = `${redirectUrl}${loginSessionToken}`;
+export const appUrlLoginScheme = "iologin";
--- a/src/routers/public.ts
+++ b/src/routers/public.ts
@@ -12,6 +12,7 @@ import * as zlib from "zlib";
 import { assetsFolder, ioDevServerConfig } from "../config";
 import { backendInfo } from "../payloads/backend";
 import {
+  appUrlLoginScheme,
   errorRedirectUrl,
   loginLolliPopRedirect,
   loginSessionToken,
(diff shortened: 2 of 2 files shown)


src/routers/features/featureFlagUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+        pipe(
+          Object.keys(osPerDevice),
+          A.findFirst(k => userAgent.includes(k)),
+          O.map(a => osPerDevice[a as keyof typeof osPerDevice])
+        )
+    )
+  );
+
+export const isFeatureFlagWithMinVersionEnabled = (
+  featureFlag: FeatureFlagWithMinAppVersion,
```
shadowsheep1, line 58:
```suggestion
  featureFlag: FeatureFlagWithMinAppVersion<Config>,
```

If we make `FeatureFlagWithMinAppVersion` generic we have to adapt this like above or below.

```suggestion
  featureFlag: FeatureFlagWithMinAppVersion<BackendStatus["config"]>,
```

src/routers/features/featureFlagUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+type FeatureFlagWithMinAppVersion = Extract<
+  keyof BackendStatus["config"],
+  {
+    [K in keyof BackendStatus["config"]]: BackendStatus["config"][K] extends
+      | { min_app_version?: VersionPerPlatform }
+      | undefined
+      ? K
+      : never;
+  }[keyof BackendStatus["config"]]
+>;
```
shadowsheep1, line 30:
```suggestion
type FeatureFlagWithMinAppVersion<T> = Extract<
  keyof T,
  {
    [K in keyof T]: T[K] extends
      | { min_app_version?: VersionPerPlatform }
      | undefined
      ? K
      : never;
  }[keyof T]
>;
```

We could make it generic, what do you think?

src/routers/features/featureFlagUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+  pipe(
+    O.fromNullable(backendStatus.config[featureFlag]?.min_app_version),
+    O.fold(
+      () => false,
+      (min_app_version: VersionPerPlatform) =>
+        pipe(
+          getOsFromUserAgent(reqHeaders),
+          O.fold(
+            () => false,
+            os => isVersionValidAndActive(min_app_version[os])
```
shadowsheep1, line 70: Here you're only checking if the string is a `SemVer` != from `0`s. You are not able to discriminate if the version of the FF is active for the current client that made the request? Am I correct? 

For example. If the feature flag is set to `1.2.3.5` and the request is done by an Android device running an IO-APP version `1.2.3.4`, the feature should be disabled because `1.2.3.5` is greater than `1.2.3.4`.  To add this check we have to pass the App version from the client to the backend in some way.

shadowsheep1 (changes requested):
I've just add some suggestion and on question to address. Tell me what do you think ;)

shadowsheep1 (approved):
LGTM

PR #246, opened 2023-05-09
[IOCOM-180] Service creation refactoring

Vangaorth:
## Short description
This PR refactors the Service CRUD in order to properly work with the IoDevServerConfig

- `src/payloads/services/index` has been moved into `src/payloads/services/factory`, grouping every service creation logic
- `src/payloads/services/special` has been moved and split into `src/payloads/services/special/:id/factory:Id`, in order to divide by service feature
- `src/persistence/services.ts` has been created in order to operate as a repository for services (following the same concept used for messages in `src/persistence/messages.ts`). In this way, services can be created using a custom _io-api-dev-server_ configuration during tests (which was not possible before)
- `src/routers.service.ts` has been edited in order to rely on `src/persistence/services.ts` to retrieve services' data and it does not longer generate and keep the service list
- other files that depended on services' data retrieval now use `src/persistence/services.ts`
- tests that rely on services have been updated in order to initialise the service DB before running and emptying it afterwards
- `src/types/config.ts`: services configuration has its own type (existing configurations have been updated accordingly)

## How to test
There are multiple ways by which this PR can be tested. Here are some examples:
- compare `service` endpoints' outputs with the previous implementation. Apart from random generated data, they should be the same
- run E2E on _io-app_ application. They should all succeed
- manually test _io-app_ flows that require services' data. They should all succeed


src/utils/variables.ts
```diff
@@ -170,23 +170,6 @@ en:
         action: "ioit://cgn-details/categories-merchant/cultureAndEntertainment"
 ---`;
 
-export const frontMatter1CTASiciliaVola = `---
```
CrisTofani, line 173: Why are we removing this?

Vangaorth: It has been moved under `src/payloads/services/special/siciliaVola/factorySiciliaVolaService.ts`. In this way, it is grouped under its own feature (since it is used only during service creation) and `src/utils/variables.ts` does not grow too much

CrisTofani: 👍🏻 lost it in the review

src/persistence/services.ts
```diff
@@ -0,0 +1,165 @@
...
+  servicePreferences.clear();
+};
+
+const getLocalServices = () => {
+  const clonedLocalServices: Readonly<ServicePublic>[] = [];
+  localServices.forEach(localService =>
+    clonedLocalServices.push(Object.assign({}, localService))
+  );
+  return clonedLocalServices;
+};
```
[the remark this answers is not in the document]
Vangaorth: Nice! Applied

src/persistence/services.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  return specialServices.find(service => service.service_id === serviceId);
+};
+
+const getSummaries = (
+  excludeSpecialServices: boolean = false
+): ReadonlyArray<ServiceSummary> => {
+  const services = excludeSpecialServices
+    ? localServices.concat(nationalServices)
+    : localServices.concat(nationalServices, specialServices);
```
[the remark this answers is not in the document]
Vangaorth: Nice! Applied

hevelius (approved):
LGTM. Tested using iOS simulator by performing various service related actions. I also tried running the e2 test and they resolved correctly.

CrisTofani (approved):
Lgtm

PR #248, opened 2023-05-11
[IOCOM-208] Migrate to ESLint and restore Prettier

Vangaorth:
This PR depends on #246

## Short description
This PR migrates from TSLint (which was not working anymore) to ESLint
It also updates Prettier to its latest version and restores/adds it to the CI

There are a lot of changes in this PR. They have been done to support ESLint, remove TSLint and fix all lint errors.
Lint warnings have been left for future developments to solve

## How to test
Run `yarn prettify` and `yarn prettier:check` to verify that prettier is working
Run `yarn lint` and check that it has an output with warnings but no errors


package.json
```diff
@@ -106,17 +107,23 @@
...
     "@types/sha256": "^0.2.0",
     "@types/supertest": "^2.0.8",
     "@types/xml2js": "^0.4.11",
+    "@typescript-eslint/eslint-plugin": "^5.59.5",
+    "@typescript-eslint/parser": "^5.59.5",
+    "eslint": "^8.40.0",
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-functional": "^5.0.8",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-react": "^7.32.2",
```
adelloste, line 119: This dependency should not be necessary, can we remove it?

```suggestion
```

Vangaorth: Nice catch! Removed

PR #252, opened 2023-05-17
[IOBP-46] Remove Bonus Vacanze list payload

mastro993:
## Short description
This PR removes the payload from the Bonus Vacanze activations list

- `src/routers/bonus-vacanze/index.ts`: replaced `bonus/vacanze/activations` payload with empty array

## How to test
Tests should solve correctly. 
With the IO app, no bonus data should be visible in the wallet section.


src/routers/features/bonus-vacanze/index.ts
```diff
@@ -73,7 +73,7 @@ const addPrefix = (path: string) => addApiV1Prefix(`/bonus/vacanze${path}`);
 // the authenticated user or by any between his family member
 addHandler(bonusVacanze, "get", addPrefix(`/activations`), (_, res) => {
   res.json({
-    items: aLotOfBonus.map(b => ({ id: b.id, is_applicant: true }))
+    items: [] // aLotOfBonus.map(b => ({ id: b.id, is_applicant: true }))
```
Vangaorth, line 76: Leaving commented code may be a potential increase in uncleanness of the codebase. 

What do you think about deleting the entire commented code? At any time, it can be retrieved using repository's history or, if it is planned to be restored at any future time, maybe it is better to create a Jira ticked and replace the comment with the ticket's link (and a short description)

Vangaorth (changes requested):

src/routers/features/bonus-vacanze/index.ts
mastro993: Fixed in https://github.com/pagopa/io-dev-api-server/pull/252/commits/dff1402adc0ce97d2db054e8eb4d3ca6f5e7b14d

Vangaorth (approved):
LGTM!

PR #256, opened 2023-05-23
[IOBP] Fix lint errors and warnings

mastro993:
Depends on https://github.com/pagopa/io-dev-api-server/pull/254

## Short description
This PR fixes or mutes ESLint errors and warnings for IOBP code branches.


src/payloads/payload.ts
```diff
@@ -334,6 +334,7 @@ export const getPspList = () => {
 
 // eslint-disable-next-line: readonly-array
 export const getPaymentsArray = (): paymentItem[] => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const localStorage = {} as any;
```
Vangaorth, line 338: Should not this be `as Storage` instead of `as any` (and consequently, with no `@typescript-eslint/no-explicit-any`)?

src/payloads/payload.ts
```diff
@@ -368,7 +370,8 @@ export const setPayment = (
       p => p.idTransaction === idTransaction
     );
     if (alreadyExist[0] === undefined || alreadyExist.length < 1) {
-      payments.push(payment);
+      payments = [...payments, payment];
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const localStorage = {} as any;
```
Vangaorth, line 375: Should not this be `as Storage` instead of `as any` (and consequently, with no `@typescript-eslint/no-explicit-any`)?

Vangaorth (changes requested):

src/payloads/payload.ts
mastro993: Fixed in https://github.com/pagopa/io-dev-api-server/pull/256/commits/213c2ddaaa57fa83fe7640ac643ce42f371acecd

src/payloads/payload.ts
Vangaorth: Can we also remove the `// eslint-disable-next-line`?

src/payloads/payload.ts
Vangaorth: Can we also remove the `// eslint-disable-next-line`?

mastro993: Whoops... of course. 😆
See https://github.com/pagopa/io-dev-api-server/pull/256/commits/78538b42841c4c16234074c9e749ca3a8743d8da

Vangaorth (approved):

Vangaorth:
LGTM!

PR #261, opened 2023-05-31
chore: Fix no error on empty "signature-input" header

shadowsheep1:
## Short description
This PR return an error when the `signature-input` header is empty.

See commit list

## How to test
Try the lollipop playground with IO-APP by providing an empty string for `signature-input` header.

Using Proxyman you can do that by scripting over this url: http: //localhost:3000/first-lollipop/sign

```js
async function onRequest(context, url, request) {
  // console.log(request);
  console.log(url);

  // Update or Add new headers
  //delete request.headers.signature;
  //delete request.headers["signature-input"];
  request.headers["signature-input"] = ""
  //request.headers["signature"] = ""
  //request.headers["content-digest"] = "foo"


  // Done
  return request;
}
```

### DIFF
--- a/src/routers/features/lollipop/index.ts
+++ b/src/routers/features/lollipop/index.ts
@@ -1,16 +1,14 @@
 import { pipe } from "fp-ts/lib/function";
 import * as O from "fp-ts/lib/Option";
 import * as TE from "fp-ts/lib/TaskEither";
+import * as E from "fp-ts/lib/Either";
 import * as T from "fp-ts/lib/Task";
 import * as B from "fp-ts/lib/boolean";
 import * as jose from "jose";
 import { Request, Response, Router } from "express";
 import { verifySignatureHeader } from "@mattrglobal/http-signatures";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import { getProblemJson } from "../../../payloads/error";
-/**
- * this router serves lollipop API
- */
-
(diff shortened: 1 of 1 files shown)


src/routers/features/lollipop/index.ts
```diff
@@ -55,33 +55,43 @@ const toTaskError = (
 
 addHandler(lollipopRouter, "post", "/first-lollipop/sign", async (req, res) =>
   pipe(
-    getPublicKey(),
-    O.fromNullable,
-    O.foldW(
-      () => toTaskError(res, 500, "Public key not found"),
-      publicKey =>
+    req.headers["signature-input"] !== "",
```
Vangaorth, line 58: What if we move the logic to a dedicated function, so that the pipeline becomes
```
pipe(
  req.headers["signature-input"],
  isNotEmpty,  // This is the extracted function that does inputStr !== ""
  B.foldW(
  ...
  )
)
```

Vangaorth (changes requested):

src/routers/features/lollipop/index.ts
shadowsheep1: Good point! Let's use `NonEmptyString.decode`, what do you think?

Vangaorth: Good idea!

Vangaorth (approved):
LGTM!

PR #262, opened 2023-06-01
[IOCOM-253,IOCOM-258] Lollipop middleware and third party message endpoints with lollipop

Vangaorth:
## Short description
This PR refactors lollipop checks and adds the new lollipop middleware.

- Lollipop logic in `src/routers/features/lollipop/index.ts` has been moved to `src/middleware/lollipopMiddleware.ts`
- The following endpoints use the lollipop middleware:
  - message precondition
  - third party message
  - third party message attachments
  - lollipop playground
- global configuration to disable every lollipop middleware in `/src/config.ts`
- README.md updated with instructions on how to use lollipop

## How to test
- With the global lollipop configuration enabled, check that the previous endpoints do not work with improper headers.
- Repeat the same test but with global lollipop configuration disabled (endpoints shall work)
- Using the latest master deploy of io-app (with lollipop enabled globally), check that, for any PN message:
  - preconditions are retrieved
  - third party message is retrieved
  - attachments are retrieved

### DIFF
--- a/src/middleware/lollipopMiddleware.ts
+++ b/src/middleware/lollipopMiddleware.ts
@@ -26,7 +26,6 @@ export const isLollipopConfigEnabled = () =>
 export const lollipopMiddleware =
   (nextMiddleware: (embeddedRequest: Request, _: Response) => void) =>
   async (request: Request, response: Response) => {
-    console.log(`MIDDLEWARE`);
     pipe(
       isLollipopConfigEnabled(),
       B.fold(
--- a/src/middleware/lollipopMiddleware.ts
+++ b/src/middleware/lollipopMiddleware.ts
@@ -51,8 +51,8 @@ export const lollipopMiddleware =
     );
   };
 
-const verifyLollipopSignatureHeader = async (req: Request, _: Response) =>
-  await pipe(
+const verifyLollipopSignatureHeader = (req: Request, _: Response) =>
+  pipe(
     req.headers["signature-input"],
     NonEmptyString.decode,
     E.foldW(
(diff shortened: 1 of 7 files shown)


src/middleware/lollipopMiddleware.ts
```diff
@@ -0,0 +1,131 @@
...
+          pipe(
+            await verifyLollipopSignatureHeader(request, response),
+            E.foldW(
+              error => response.status(error.code).send(error.problemJson),
+              _ => nextMiddleware(request, response)
+            )
+          )
+      )
+    );
+  };
```
shadowsheep1, line 44: Here we could avoid `async/await` with a more `fp-ts` idiomization... 

```suggestion
export const lollipopMiddleware =
  (nextMiddleware: (embeddedRequest: Request, _: Response) => void) =>
  (request: Request, response: Response) => {
    pipe(
      isLollipopConfigEnabled(),
      B.fold(
        () => nextMiddleware(request, response),
        () =>
          pipe(
            TE.tryCatch(
              () => verifyLollipopSignatureHeader(request, response),
              _ => _ as Error
            ),
            TE.map(verificationResult =>
              pipe(
                verificationResult,
                E.foldW(
                  error => response.status(error.code).send(error.problemJson),
                  _ => nextMiddleware(request, response)
                )
              )
            )
          )()
      )
    );
  };
```
Is a bit more verbose, as usual... 😊. What do you think?

Vangaorth: Yeah I prefer the one you are suggesting, let's adopt it

src/middleware/lollipopMiddleware.ts
```diff
@@ -0,0 +1,139 @@
...
+                )
+              )
+            )
+          )()
+      )
+    );
+  };
+
+const verifyLollipopSignatureHeader = async (req: Request, _: Response) =>
+  await pipe(
```
shadowsheep1, line 55: We can now remove `async/await` here too, without any further modification. Am I right?

Vangaorth: Totally right, corrected

shadowsheep1 (approved):
LGTM

PR #265, opened 2023-06-09
[IOPID-313] FastLogin

SabPicariello:
## Short description
This PR adds the fast login logic to dev server

- src/features/fastLogin/fastLoginHandler.ts: handler for fast login request.
- src/features/fastLogin/nonceHandler.ts: handler for nonce request.
- src/middleware/fastLoginMiddleware.ts: middleware that returns 401 on every call if the sessionToken is not valid.
- src/middleware/lollipopMiddleware.ts: added error 403 to lollipop middleware. The error is triggered only if assertionRef is configured to have an expiration date.
- src/payloads/login.ts: removed hard coded info about sessionToken.
- src/persistence/appInfo.ts: added os information.
- src/persistence/lollipop.ts: added logic to handle assertionRef expiration.
- src/persistence/sessionInfo.ts: added info about session and its management.
- src/routers/__tests__/server.test.ts: updated test.
- src/routers/features/fastLogin/index.ts: added router for fast login and nonce.
- src/routers/features/featureFlagUtils.ts: updated.
- src/routers/public.ts:  added some clear function to logout, optimized code.
- src/server.ts: added FastLogin router and middleware.
- src/types/config.ts: added configurations for assertionRef and token session validity.
- src/utils/date.ts: added a util function.

## How to test
- Use dev server as always and verify that everything works as expected.
- in config.js, set an expiration time for assertionRef and test lollipop calls. It should work until it expires.
- in config.js, set an expiration time for assertionRef and sessionToken, enable the feature flag and make sure the app sends the right header value for fast login. It should follow the fast login spec.
- with every fast login feature configured, try to perform a standard login by setting the login header in the app. It should work as usual.
- all incorrect configurations (for example an app that requires a FastLogin but the ff or session duration time has not been set on the server), should - rightly - give an error


config/config.json
```diff
@@ -3,5 +3,6 @@
     "attrs": {
       "name": "Gian Maria"
     }
+
```
shadowsheep1, line 6: Why do we have an empty space here? ;)

src/persistence/sessionInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+  Object.keys(loginSessionTokenInfo).forEach(key => {
+    loginSessionTokenInfo[key as keyof LoginSessionTokenInfo] = undefined;
+  });
+
+const setLoginSessionToken = (newToken: string) => {
+  loginSessionTokenInfo.loginSessionToken = newToken;
+  loginSessionTokenInfo.instantiationDate = new Date();
+};
+
+const generateNewLoginToken = () => faker.random.alphaNumeric(11).toUpperCase();
```
shadowsheep1, line 38: And what about third party tokens? Maybe we should regenerate them too, just to have a quick feedback they've changed too? What do you think?

shadowsheep1 (changes requested):
I've requested some minor changes.

src/utils/date.ts
```diff
@@ -28,3 +28,6 @@ export class DateFromISOStringType extends t.Type<Date, string, unknown> {
 
 export const DateFromISOString: DateFromISOStringType =
   new DateFromISOStringType();
+
+export const getDateMsDifference = (firstDate: Date, secondDate: Date) =>
+  firstDate.getTime() - secondDate.getTime();
```
shadowsheep1, line 33: I'd use the absolute value to be firstDate/secondDate temporal reference agnostic... what do you think?

shadowsheep1 (changes requested):

src/utils/date.ts
SabPicariello: you are absolutely right :+1:

shadowsheep1 (approved):
LGTM

PR #267, opened 2023-06-16
[IOBP-68] IDPay transaction authorization failures

mastro993:
## Short description
This PR adds the ability to test IDPay transaction failures.
Each failure type is mapped from a specific digit extracted from the requested `trxCode` as follows:

| Code | Status | Failure |
| ---- | ------ | ----------- |
| 1    | 404    | `PAYMENT_NOT_FOUND_EXPIRED` |
| 2    | 403    | `PAYMENT_USER_NOT_VALID` |
| 3    | 400    | `PAYMENT_STATUS_NOT_VALID` |
| 4    | 403    | `PAYMENT_BUDGET_EXHAUSTED` |
| 5    | 403    | `PAYMENT_GENERIC_REJECTED` |
| 6    | 429    | `PAYMENT_TOO_MANY_REQUESTS` |
| 7    | 500    | `PAYMENT_GENERIC_ERROR` |

Code is extracted from the `trxCode` using the mask "XXXXXXAB", where:
- `X`: ignored character
- `A` is the digit for the failure to expect during the pre-authorization
- `B` is the digit for the failure to expect during the authorization.

Some examples of trxCode used during the authorization process in the IO app: 
- `00000001`: pre-authorization success but 404 and `PAYMENT_NOT_FOUND_EXPIRED` failure during authorization. 
- `00000030`: pre-authorization failure with 400 and `PAYMENT_STATUS_NOT_VALID`.
- `00000000`: no failure

## How to test
Tests should solve correctly


src/routers/features/idpay/payment.ts
```diff
@@ -1,21 +1,75 @@
...
     O.fromNullable,
-    O.chain(putPreAuthPaymentResponse),
+    O.map(trxCode => trxCode[trxCode.length - 2]),
+    O.map(parseInt),
+    O.filter(code => code <= 7),
+    O.chain(code => O.fromNullable(codeToFailure[code])),
     O.fold(
-      () => res.status(404).json(getIdPayError(404)),
-      data => res.status(200).json(data)
+      () => res.status(200).json(generateRandomAuthPaymentResponseDTO()),
```
forrest57, line 70: if the pre auth were to pass, but the auth should fail, since the TRXcode in `generateRandomAuth...` is a random string, usually a generic error happens but it could be anything. 
I suggest adding a non-mandatory parameter for the TRXcode, so that it can be passed back in the pre auth but not in the auth.
TLDR: `000000001` passed in the auth -> pre-auth passes but auth fails with generic error since the TRX returned and passed to the BE is along the lines of `(4ve?b>{|`, while it should fail with a 404 error

mastro993: fixed in https://github.com/pagopa/io-dev-api-server/pull/267/commits/c106018c09d52cb1f1d1467115d4f7deda31073c

forrest57 (approved):

forrest57:
LGTM!

PR #271, opened 2023-06-21
[IOPID-355] Updated Readme and config for FL

SabPicariello:
## Short description
Updated Readme and config for FL

- README.MD: added section for FL.
- src/config.ts: added default value for session TTL.
- src/persistence/sessionInfo.ts: added an error in console to be more explicit in case it occurs.

## How to test
Enable fast login FF without manually configure a TTL and try to run IO app with a brach that includes FL feature.


shadowsheep1 (approved):
LGTM

PR #277, opened 2023-06-28
[IOPID-286] Add fast-login api Open API specs

shadowsheep1:
## Short description
This PR adds the `fast-login` endpoint Open API specs, the models generation and use them in code.

* [package.json](https://github.com/pagopa/io-dev-api-server/blob/d543e6ddecf4b5996b21d8a1bb367db2778b049c/package.json): Adds fast login models generation.
* [src/middleware/lollipopMiddleware.ts](https://github.com/pagopa/io-dev-api-server/blob/d543e6ddecf4b5996b21d8a1bb367db2778b049c/src%2Fmiddleware%2FlollipopMiddleware.ts): Returns `403` instead of `500` in case of missing key.
* [src/routers/features/fastLogin/index.ts](https://github.com/pagopa/io-dev-api-server/blob/d543e6ddecf4b5996b21d8a1bb367db2778b049c/src%2Frouters%2Ffeatures%2FfastLogin%2Findex.ts): Uses `FastLoginResponse` type.

## How to test
Enable the fast login feature for the IO-APP by setting in `src/payloads/backend.ts`:
```
fastLogin: {
      min_app_version: {
        ios: "1.2.3",
        android: "1.2.3"
      }
    },
```
and run the app against this branch. If you login in app and then you stop and restart the server, the app should do a log-out because it receives a `403` error from the `fast-login` endpoint.


SabPicariello (approved):
LGTM

PR #282, opened 2023-07-14
[IOCOM-363] PN Opt In Message

Vangaorth:
## Short description
This PR adds support for the PN opt in message

Code has been edited to generate both PN opt in service and PN opt in message

## How to test
Change the server configuration to enabled pn and generate the opt in message. Either use io-app or cUrl/postman to check both message and service

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "api_cgn": "https://raw.githubusercontent.com/pagopa/io-backend/v12.6.1-RELEASE/api_cgn.yaml",
   "api_cgn_merchants": "https://raw.githubusercontent.com/pagopa/io-backend/v12.6.1-RELEASE/api_cgn_operator_search.yaml",
   "api_cgn_geo": "https://raw.githubusercontent.com/pagopa/io-backend/here_geoapi_integration/api_geo.yaml",
-  "content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.25/definitions.yml",
+  "content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.26/definitions.yml",
   "api_pagopa_walletv2": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.25/bonus/specs/bpd/pm/walletv2.json",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "api_cgn": "https://raw.githubusercontent.com/pagopa/io-backend/v12.6.1-RELEASE/api_cgn.yaml",
   "api_cgn_merchants": "https://raw.githubusercontent.com/pagopa/io-backend/v12.6.1-RELEASE/api_cgn_operator_search.yaml",
   "api_cgn_geo": "https://raw.githubusercontent.com/pagopa/io-backend/here_geoapi_integration/api_geo.yaml",
-  "content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.25/definitions.yml",
+  "content_specs": "https://raw.githubusercontent.com/pagopa/io-services-metadata/1.0.26/definitions.yml",
```
adelloste, line 13: Should we align the specs?

src/payloads/services/special/pn/factoryPn.ts
```diff
@@ -10,15 +10,43 @@ import { ServiceId } from "../../../../../generated/definitions/backend/ServiceI
...
+export const pnOptInCTA = `---
+it:
+    cta_1: 
+        text: "Attiva il servizio"
+        action: "ioit://services/service-detail?serviceId=servicePN&activate=true"
+en:
+    cta_1: 
+        text: "Turn service on"
+        action: "ioit://services/service-detail?serviceId=servicePN&activate=true"
+---`;
```
adelloste, line 24: Could it make sense to move this inside `src/utils/variables.ts`?

adelloste (commented):
I left some comments

src/payloads/services/special/pn/factoryPn.ts
Vangaorth: It depends if we want to keep all under a single file or if we want to move each piece of code under its own domain. What do you think about it?

package.json
Vangaorth: Yep, done it 7a390f0

adelloste (approved):

PR #286, opened 2023-07-19
feat: [IOBP-132] New wallet onboarding web page mock

Hantex9:
## Short description
This PR adds a new webpage to somulate the wallet onboarding page with the possibility to choose which outcome we want emit;

- Added `wallet_onboarding.html` page that contains a select with a list of possible outcomes and a button to propagate the selected outcome;
- Added a post API to simulate the onboarding start which gives as a response the `walletId` and `redirectUrl` which will be used into WebView from the app;


mastro993 (approved):
LGTM

PR #293, opened 2023-08-11
Fix: IDPay initiatives status by wallet id

mastro993:
## Short description
This PR fixes the `/wallet/instrument/{walletId}/initiatives` endpoint, which returns the initiatives status for a payment method.

- `getInitiativeWithInstrumentResponse` now returns the status of all initiatives and not only for the active initiatives
-  `getWalletById` now has an `enabledFunctions` filters to exclude wallets without BPD capabilities

## How to test
Tests should solve correctly


Hantex9 (approved):
LGTM!

PR #297, opened 2023-09-01
chore: [IOPID-714,IOPID-670]: Discriminate currentProfile based on auth provider

shadowsheep1:
## Short description
This PR allows the app to retrieve the profile data based on the authentication provider ("spid"/"cie") passed through a custom header (`x-pagopa-idp-id`) during the login request.

## How to test
Whit this https://github.com/pagopa/io-app/pull/4948, you can test that the dev server uses the EIC or the SPID user profile data based on the login you are attempting to do. If you start an EIC login flow, after the login you will have the `cieProfile` or `cieProfileFirstOnboarding` data. Otherwise you  will have the `spidProfile` or `spidProfileFirstOnboarding`one.


src/persistence/sessionInfo.ts
```diff
@@ -15,14 +16,19 @@ type LoginSessionTokenInfo = {
...
   loginType: LoginEnum | undefined;
+  authenticationProvider: AuthenticationProvider | undefined;
 };
 
 const loginSessionTokenInfo: LoginSessionTokenInfo = {
   loginSessionToken: undefined,
   instantiationDate: undefined,
-  loginType: undefined
+  loginType: undefined,
+  authenticationProvider: undefined
```
Vangaorth, line 26: This way, we are changing the default behaviour for which, upon first startup, `io-dev-api-server` already has the capability of generating data with a valid profile.
Can we provide a default value instead of `undefined`, like
```
  loginType: LoginEnum.standard,
  authenticationProvider: "spid"
```
?

Maybe we can also change the type so it is never `undefined`?

Vangaorth (changes requested):

src/routers/profile.ts
```diff
@@ -12,15 +13,17 @@ import {
...
 import { mockUserMetadata } from "../payloads/userMetadata";
 import { getRandomValue } from "../utils/random";
 import { addApiV1Prefix } from "../utils/strings";
 import { validatePayload } from "../utils/validator";
+import { getAuthenticationProvider } from "../persistence/sessionInfo";
+import { InitializedProfile } from "../../generated/definitions/backend/InitializedProfile";
 
 // eslint-disable-next-line functional/no-let
-export let currentProfile = { ...profile };
+export let currentProfile: InitializedProfile = {} as InitializedProfile;
```
Vangaorth, line 26: Should we also not export this? If we do not do it, the `getCurrentProfile` method can be ignored and the profile is accessed without the proper flow.

Vangaorth (changes requested):

src/routers/profile.ts
shadowsheep1: Good point. Thank you.

shadowsheep1: Fixed.

PR #300, opened 2023-09-11
[IOCOM-533] Add `min_app_version` to `pn`

adelloste:
## Short description
This PR adds `min_app_version` field to `pn`.


src/payloads/backend.ts
```diff
@@ -84,6 +84,10 @@ export const backendStatus: BackendStatus = {
     },
     pn: {
       enabled: true,
+      min_app_version: {
+        ios: "1.2.3",
```
Vangaorth, line 88: Should we set this to the current min app version? `2.35.0.1`

Vangaorth (changes requested):

Vangaorth (approved):
LGTM

PR #301, opened 2023-09-12
[chore] PN configuration disabled by default

Vangaorth:
## Short description
This PR disables the active PN configuration

- config: changed in order to not generate any PN message or service by default

## How to test
Run the server instance and check that there are no PN messages no services


shadowsheep1 (approved):
LGTM

PR #307, opened 2023-10-03
[IOCOM-577] Removal of the `category` property from `PaymentProblemJson`

adelloste:
## Short description
This PR removes the `category` property from `PaymentProblemJson`.

- updated definitions
- removed the category property from the generated payload

## How to test
Generate a payment that contains an error or that is already paid and check that the verify endpoint returns a payload without category.


Vangaorth (approved):
LGTM

PR #317, opened 2023-11-07
[IOPID-991] add new openapi configuration

Ladirico:
## Short description
Add configuration of CDU flow from[ this branch row](https://github.com/pagopa/io-backend/tree/IOPID-1045-1050)


shadowsheep1 (approved):
LGTM

PR #319, opened 2023-11-14
chore: [IOBP-406] adapted new wallet details apis

Hantex9:
## Short description
This PR adapts some new wallet details informations and apis in order to be compliant at the new payment gateway

- Updated the swagger definition;
- Added an handling into `wallet_onboarding.html` that creates a mock payment method based on the user selection;
- Adapted the `generateUserWallet` method in order to have all the required information requested by the new wallet payment apis;
- Added a new `getWalletTypeFromPaymentMethodId` method that will get the wallet type from the payment method id selected

## How to test
Checkout this PR and test it with the following PR: https://github.com/pagopa/io-app/pull/5189 - You should be able to onboard all the available payment methods (in this case 2: CARD and PAYPAL)


src/routers/features/walletV3/router.ts
```diff
@@ -3,9 +3,10 @@ import { addHandler, SupportedMethod } from "../../../payloads/response";
 
 export const walletV3Router = Router();
 
-export const addWalletV3Prefix = (path: string) => `/wallets${path}`;
+export const addWalletV3Prefix = (path: string) =>
+  `/payment-wallet/v1/wallets${path}`;
 export const addPaymentMethodsPrefix = (path: string) =>
-  `/payment-methods${path}`;
+  `/payment-wallet/v1/payment-methods${path}`;
```
mastro993, line 9: I think the prefix should be only one: `payment-wallet/v1`. Everything else it's part of the single handler.
What do you think?

mastro993 (changes requested):

src/routers/features/walletV3/router.ts
Hantex9: I agree with you to add `/payment-wallet/v1` as common prefix, but I think that exporting it as a constant and appending to these two methods is better. Since these two methods are used to append just the prefix for each service (`/wallets` and `/payment-methods`) and the handler will do the rest.

Check it out in this commit what I mean: 0b0cfb3e31aa92765718dc05745d79bf408555a2

mastro993 (approved):

PR #323, opened 2023-11-27
[IOCOM-686] Add message generation with remote content

adelloste:
## Short description
This PR allows to generate messages with remote content.

- `src/config`: added configuration for messages with remote content
- `src/populate-persistence.ts`: added generation of messages with remote content (saved in-memory)
- `src/features/messages/persistence/messagesPayload.ts`: added `withRemoteContent` function that adds remote content to the message

## How to test
Run the server instance with different configurations and check that the payload returned by the `/messages/:id` and `/third-party-messages/:id` endpoints is correct

### DIFF
--- a/src/features/messages/persistence/messagesPayload.ts
+++ b/src/features/messages/persistence/messagesPayload.ts
@@ -73,7 +73,8 @@ export const withRemoteContent = (
     ...message.content,
     third_party_data: {
       ...message.content.third_party_data,
-      id: message.id as NonEmptyString
+      id: message.id as NonEmptyString,
+      has_attachments: true
     }
   },
   third_party_message: {
--- a/src/features/messages/persistence/messagesPayload.ts
+++ b/src/features/messages/persistence/messagesPayload.ts
@@ -63,24 +63,6 @@ export const withDueDate = (
   content: { ...message.content, due_date: dueDate }
 });
 
-export const withRemoteAttachments = (
-  message: CreatedMessageWithContent,
-  attachmentCount: number
-): ThirdPartyMessageWithContent => ({
-  ...message,
-  content: {
-    ...message.content,
-    third_party_data: {
-      ...message.content.third_party_data,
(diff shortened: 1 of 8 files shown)


src/features/messages/persistence/messagesPayload.ts
```diff
@@ -61,21 +63,33 @@ export const withDueDate = (
...
-  attachmentCount: number
+  markdown: string
 ): ThirdPartyMessageWithContent => ({
   ...message,
   content: {
     ...message.content,
     third_party_data: {
       ...message.content.third_party_data,
-      id: message.id as NonEmptyString,
-      has_attachments: true
```
Vangaorth, line 74: This flag is sent by the real backend, we should keep it and assign a proper value to it. 
We also need to add the `has_remote_content: boolean` flag.
What do you think?

src/features/messages/services/messagesService.ts
```diff
@@ -83,7 +83,7 @@ const getPublicMessages = (
       B.fold(
         () => ({}),
         () => {
-          const { content, is_read, is_archived, has_attachments } =
```
Vangaorth, line 86: See previous comment on keeping this value, also with the new `has_remote_content`

src/features/messages/services/messagesService.ts
```diff
@@ -93,7 +93,6 @@ const getPublicMessages = (
             category: getMessageCategory(message),
             is_read,
             is_archived,
-            has_attachments,
```
Vangaorth, line 96: See previous comment on keeping this value, also with the new `has_remote_content`

Vangaorth (changes requested):

src/features/messages/persistence/messagesPayload.ts
adelloste: Restored `has_attachments` flag

Vangaorth (approved):
LGTM

PR #330, opened 2023-12-05
fix: add default values for validated and taken fields

shadowsheep1:
## Short description
This PR add default values for `is_email_validated` and `is_email_already_taken`.


Hantex9 (approved):
LGTM!

PR #333, opened 2023-12-14
chore: set user email as not mandatory

shadowsheep1:
## Short description
This PR makes email user field as optional.


Ladirico (approved):
LGTM

PR #345, opened 2024-02-01
[IOPID-1448] chore: add 409 management

shadowsheep1:
## Short description
This PR adds the `409` error in the profile `POST` if the version is not the actual one based on  [io-functions-app UpdateProfile/handler.ts#L135](https://github.com/pagopa/io-functions-app/blob/9ee5f40de8ed8d8505c025bc1c3816e7fc0653da/UpdateProfile/handler.ts#L135)

## How to test
Test IO app against this branch. All must work fine.


Ladirico (approved):
LGTM!

PR #346, opened 2024-02-01
[SFEQS-2100] Update `Firma con IO` specs

hevelius:
## Short description
This PR updates Firma con IO definitions to avoid decoder fails when status response code is 401.

- Update package.json

## How to test
Static test should be enough. As a plus try a signing flow.


LazyAfternoons (approved):
LGTM

PR #349, opened 2024-02-13
[IOCOM-1091] Add a message endpoint to create a new message

Vangaorth:
## Short description
This PR adds a new endpoint to create a new message while the dev-server is running.

- message creation in `MessagesService`
- message in-memory storage in `MessagesDatabase`
- message creation endpoint in `MessagesRouter`

## How to test
Call the `POST message` endpoint and check that the message is correctly created and retrieved using the `messages` endpoint

### DIFF
--- a/assets/html/homeDashboard.html
+++ b/assets/html/homeDashboard.html
@@ -7,5 +7,6 @@
     <body> 
         <include>assets/html/wallet2_config.html</include>
         <include>assets/html/mail.html</include>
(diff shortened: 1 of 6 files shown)


adelloste:
It might be useful to add this call in the `homeDashboard.html`, what do you think?

Vangaorth:
> It might be useful to add this api call in the `homeDashboard.html`, what do you think?

Agreed, added

adelloste (approved):

PR #360, opened 2024-04-08
[IOCOM-1178] FIMS OIdC flow

Vangaorth:
## Short description
This PR adds a fake OIdC flow for FIMS, with both Relying Party and Provider support.

- relyingParty router that:
  - redirects to provider
  - handles the redirection router, checking for mandatory parameters (like nonce, state and lollipop-headers presence)
  - decodes and shows the signed idToken to display fake citizen informations
- provider router that:
  - handles the OIdC flows, with proper and mandatory parameters
  - supports both Authorisation Code flow and Implicit Code Flow, like the real provider
  - sets and validates cookies, like the real provider (even the FIMS authorizasion cookie)
  - signs and generate the idToken
  - simulates the Fast Login expiration
- history endpoint that generates and handles a fake paginated access history and simulates its export 

All of the above is configurable with the config.ts file

## How to test
Start the dev-server and type `http://localhost:3000/fims/relyingParty/1/landingPage` in your web-browser (or use something like cUrl to follow such link)


forrest57 (approved):
LGTM

PR #361, opened 2024-04-09
chore: add expired card in wallet

mastro993:
## Short description
This PR adds an expired payment method in the wallet payload


forrest57 (approved):
LGTM

PR #366, opened 2024-04-16
[IOPID-1742] Change urls for logo items form io.italia.it to io.pagopa.it

Ladirico:
## Short description
Change urls for logo items form io.italia.it to io.pagopa.it

- In list.json change urls to get the idps logos from io.italia.it to io.pagopa.it


Vangaorth (approved):
LGTM

PR #381, opened 2024-05-21
[IOCOM-1436] Services and Organizations random logos

Vangaorth:
## Short description
This PR adds random logos for services and organizations.

- Two lists of services and organizations file logos have been added
- On both endpoints (`static_contents/logos/services/service_id` and `static_contents/logos/organizations/organization_id`), on the first time, logo image is picked randomly from related file list. Subsequent requests to the same resource return the same file
- Special services (SEND and CGN) requests always return their own logo
- Enriched message response includes the organization fiscal code (in order to support fallback on message home list)

## How to test
Call both enpoints above and check that the same logo is returned for subsequent requests (apart from restarting the server instance). Check that both SEND and CGN always return their own logo


Vangaorth:
@adelloste

adelloste (approved):

PR #391, opened 2024-06-19
Fix native login on iOS phisical device

Ladirico:
## Short description
In order to allow native login on ios physical devices, this change was made by adding Darwin as the operating system. Before this change, the physical ios device was not interpreted correctly.

## How to test
In io-app .env file set NATIVE_LOGIN_ENABLED=YES
In io-dev-api-server in `src/payloads/backend.ts` set 
```
  nativeLogin: {
      min_app_version: {
        ios: "1.2.3",
        android: "1.2.3"
      }
    },
```
and run the application. Now the native login is active


shadowsheep1 (approved):
LGTM

PR #408, opened 2024-09-04
feat(backend): update definitions and add cie_id FF config

ChrisMattew:
>[!Important]
>Depends on [this PR](https://github.com/pagopa/io-services-metadata/pull/829)
- Updated definitions
- Added `cie_id` config 

## How to test
Run `yarn generate:content-definitions` and the type error in the `src/payloads/backend.ts` file will disappear!


shadowsheep1 (approved):
LGTM

PR #414, opened 2024-10-03
fix: [SIW-1702] Wrong response to the `POST` request for trial system

LazyAfternoons:
## Short description
This PR fixes the wrong response to the `POST` request endpoint for the trial system.

- Return `404` if a trial id doesn't exists in the configuration of the dev server;
- Return `201` instead of `200` when subscribing; 
- Return `409` if the user is already `SUBSCRIBED` or `ACTIVE` (tested in PROD to understand how the API is supposed to work).

## How to test
This can be tested by using `io-app` with the itw trial id which is already defined in the `config.ts` file. Open the profile section, enable the debug menu and in "Documenti su IO" section you can try to subscribe to the trial system.


src/config.ts
```diff
@@ -108,7 +108,7 @@ const defaultConfig: IoDevServerConfig = {
     paymentsCount: 1,
     paymentInvalidAfterDueDateWithValidDueDateCount: 0,
     paymentInvalidAfterDueDateWithExpiredDueDateCount: 0,
-    paymentWithValidDueDateCount: 0,
+    paymentWithValidDueDateCount: 1,
```
hevelius, line 111: Why also this value is changed?

src/config.ts
```diff
@@ -124,7 +124,7 @@ const defaultConfig: IoDevServerConfig = {
       }
     },
     withCTA: false,
-    withEUCovidCert: false,
+    withEUCovidCert: true,
```
hevelius, line 127: ditto

hevelius (changes requested):

src/config.ts
LazyAfternoons: Unwanted changes, thanks. Removed.

hevelius (approved):

PR #417, opened 2024-10-04
[IOPID-2093] Parameterisation of getSession endpoint

Ladirico:
## Short description
In order to adapt the changes in the getSession, the possibility of passing parameters was added. 
- If parameters are passed, the getSession returns the required tokens
- In the case where no parameters are passed, getSession returns all the information

## How to test
Run io-app using local environment and follow the instructions of [this PR.](https://github.com/pagopa/io-app/pull/6196)


ChrisMattew (approved):

src/payloads/session.ts
```diff
@@ -21,14 +22,35 @@ export const FIMSToken = () => {
...
+};
+
+const generateSessionTokens = (): PublicSession => tokensObject;
+
+export const generateCustomObjectSessionTokens = (
+  tokenString: string | ParsedQs | string[] | ParsedQs[]
+): Partial<PublicSession> => {
+  // If no tokenString is provided, return the entire tokensObject
+  if (!tokenString || typeof tokenString !== "string") {
+    return tokensObject;
```
shadowsheep1, line 41: Shouldn't be better here to get the entire payload from `customSession` instead of `tokensObject`. Actually they are the same but if in the future we randomize the SSO Token when we randomize the session token during a fast login refresh, the variable with fresh data'd be `customSession`. Am I right? 
Just treat `tokenObject` as the default initial value. What do you think?
```suggestion
    return customSession;
```

shadowsheep1 (changes requested):

src/payloads/session.ts
```diff
@@ -21,14 +22,35 @@ export const FIMSToken = () => {
   return mFIMSToken;
 };
 
-const generateSessionTokens = (): PublicSession => ({
+const tokensObject: PublicSession = {
```
shadowsheep1, line 25: This was a function to let all SSO token to update during a fast login refresh. Am I right? We should keep that as a function.

shadowsheep1 (changes requested):

src/payloads/session.ts
Ladirico: I applied the change, it should work now! 🚀

shadowsheep1 (approved):
LGTM

PR #418, opened 2024-10-04
[IOPLT-639] Handle the new status message payload

CrisTofani:
## Short description
This PR introduces the specs and payload on backend.ts for new status messages.


src/payloads/backend.ts
```diff
@@ -46,10 +49,6 @@ export const backendStatus: BackendStatus = {
     fims: {
       domain: `${serverUrl}/fims/provider/`,
       enabled: true,
```
CrisTofani, line 51:
```suggestion
      enabled: true,
      min_app_version: {
        android: "2.68.0.0",
        ios: "2.68.0.0"
      },
```

PR #423, opened 2024-10-16
chore: [IOBP-927] Adapt new openAPI biz-events

Hantex9:
## Short description
This PR adapts the new openAPI provided by biz events;

- Adapted the new open API of biz-events replacing from "transaction" to "notice";

## How to test
Test this PR with the following io-app PR: https://github.com/pagopa/io-app/pull/6187 in dev-env.

### DIFF
--- a/src/features/payments/persistence/notices.ts
+++ b/src/features/payments/persistence/notices.ts
@@ -82,13 +82,14 @@ const generateUserNotice = (
         .toString()
     })
   );
-  // eslint-disable-next-line functional/immutable-data
-  randomNotice.isCart = cartList.length > 1;
-  // eslint-disable-next-line functional/immutable-data
-  randomNotice.amount = cartList
-    .reduce((acc, item) => acc + Number(item.amount), 0)
-    .toString();
-  addUserNotice(randomNotice);
+  const updatedNotice = {
+    ...randomNotice,
+    isCart: cartList.length > 1,
+    amount: cartList
+      .reduce((acc, item) => acc + Number(item.amount), 0)
+      .toString()
+  };
+  addUserNotice(updatedNotice);
 
   const randomNoticeDetails: NoticeDetailResponse = {
     infoNotice: generateRandomInfoNotice(cartList),
--- a/src/features/payments/persistence/notices.ts
+++ b/src/features/payments/persistence/notices.ts
@@ -4,48 +4,45 @@ import * as O from "fp-ts/lib/Option";
 import { NoticeListItem } from "../../../../generated/definitions/pagopa/transactions/NoticeListItem";
 import { NoticeDetailResponse } from "../../../../generated/definitions/pagopa/transactions/NoticeDetailResponse";
 import { TransactionInfo } from "../../../../generated/definitions/pagopa/ecommerce/TransactionInfo";
-import { generateRandomInfoTransaction } from "../utils/transactions";
+import { generateRandomInfoNotice } from "../utils/transactions";
 import { ioDevServerConfig } from "../../../config";
 
 const mockedTaxCodes = ["1199250158", "13756881002", "262700362", "31500945"];
 
-type TransactionId = NoticeListItem["eventId"];
+type EventId = NoticeListItem["eventId"];
 
-const userTransactions = new Map<TransactionId, NoticeListItem>();
-const transactions = new Map<TransactionId, NoticeDetailResponse>();
+const userNotices = new Map<EventId, NoticeListItem>();
+const notices = new Map<EventId, NoticeDetailResponse>();
 
-const getUserTransactions = () =>
-  Array.from(userTransactions.size > 0 ? userTransactions.values() : []);
+const getUserNotices = () =>
+  Array.from(userNotices.size > 0 ? userNotices.values() : []);
 
-const getTransactionDetails = (transactionId: TransactionId) =>
+const getNoticeDetails = (eventId: EventId) =>
   pipe(
-    transactions,
+    notices,
     O.fromNullable,
-    O.chain(transactions => O.fromNullable(transactions.get(transactionId)))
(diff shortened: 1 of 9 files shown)


src/features/payments/routers/transactions.ts
```diff
@@ -3,30 +3,27 @@ import { pipe } from "fp-ts/lib/function";
...
 
-import { TransactionListWrapResponse } from "../../../../generated/definitions/pagopa/transactions/TransactionListWrapResponse";
+import { NoticeListWrapResponse } from "../../../../generated/definitions/pagopa/transactions/NoticeListWrapResponse";
 import { addTransactionHandler } from "./router";
 
 const CONTINUATION_TOKEN_HEADER = "x-continuation-token";
 
-addTransactionHandler("get", "/transactions", (req, res) => {
+addTransactionHandler("get", "/paids", (req, res) => {
   const size = req.query.size ? Number(req.query.size) : 10;
```
LeleDallas, line 12: Defined DEFAULT_SIZE like CONTINUATION_TOKEN_HEADER for better readability
`const DEFAULT_SIZE = 10;`

src/features/payments/routers/transactions.ts
```diff
@@ -3,30 +3,27 @@ import { pipe } from "fp-ts/lib/function";
...
   const offset =
     req.headers[CONTINUATION_TOKEN_HEADER] !== undefined &&
     req.headers[CONTINUATION_TOKEN_HEADER] !== "undefined"
       ? Number(req.headers[CONTINUATION_TOKEN_HEADER])
       : 0;
```
LeleDallas, line 17: That could be replaced with `isNan` checker:
``` ts
const offset = isNaN(Number(req.headers[CONTINUATION_TOKEN_HEADER])) ?
 0: Number(req.headers[CONTINUATION_TOKEN_HEADER])

File: src/features/payments/persistence/transactions.ts [hunk 3]
```diff
@@ -1,18 +1,18 @@
...
+import { NoticeListItem } from "../../../../generated/definitions/pagopa/transactions/NoticeListItem";
+import { NoticeDetailResponse } from "../../../../generated/definitions/pagopa/transactions/NoticeDetailResponse";
 import { TransactionInfo } from "../../../../generated/definitions/pagopa/ecommerce/TransactionInfo";
 import { generateRandomInfoTransaction } from "../utils/transactions";
 import { ioDevServerConfig } from "../../../config";
 
 const mockedTaxCodes = ["1199250158", "13756881002", "262700362", "31500945"];
 
-type TransactionId = TransactionListItem["transactionId"];
+type TransactionId = NoticeListItem["eventId"];
```
Ladirico, line 12: Since you are no longer using `TransactionId` can you rename this type to a more consistent name (like `eventId`), what do you think?

Hantex9: Thank you to pointing it out! I refactored the entire file replacing the word occurrencies `Transaction` with the word `Notice` (51178e7c83c603925f1560b685ce273c9af7e631)

src/features/payments/routers/transactions.ts
Hantex9: Good catch! Addressed into c1e4bb96821658c23bf909c49c6b4752122587e4

src/features/payments/routers/transactions.ts
Hantex9: Good catch! Addressed into c1e4bb96821658c23bf909c49c6b4752122587e4

src/features/payments/persistence/notices.ts
```diff
@@ -0,0 +1,121 @@
...
+        name: faker.name.fullName(),
+        taxCode: faker.random.alphaNumeric(16).toUpperCase()
+      },
+      refNumberType: "IBAN",
+      refNumberValue: faker.datatype
+        .number({ min: 100000000000, max: 999999999999 })
+        .toString()
+    })
+  );
+  // eslint-disable-next-line functional/immutable-data
```
LeleDallas, line 85: I think we we can remove the eslint comment and solve the error.
To ensure that we do not modify existing objects or arrays, we need to create new instances instead of altering the original ones

```ts
const updatedNotice = {
    ...randomNotice,
    isCart: cartList.length > 1,
    amount: cartList
      .reduce((acc, item) => acc + Number(item.amount), 0)
      .toString()
  };
  addUserNotice(updatedNotice);
```

src/features/payments/persistence/notices.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const generateUserNoticeData = () => {
+  for (
+    // eslint-disable-next-line functional/no-let
+    let i = 0;
+    i < ioDevServerConfig.features.payments.numberOfTransactions;
+    i = i + 1
+  ) {
+    generateUserNotice(faker.datatype.uuid(), i);
+  }
```
LeleDallas, line 109: We can remove the eslint comment and replace the `for` loop with a `for .. of` loop:
```ts
const generateUserNoticeData = () => {
  for (const i of Array(ioDevServerConfig.features.payments.numberOfTransactions).keys()) {
    generateUserNotice(faker.datatype.uuid(), i);
  }
};

```

LeleDallas (commented):
I believe we can remove the ESLint comments and try to solve the errors.

src/features/payments/persistence/notices.ts
Hantex9: Good idea, addressed into 59a9b9b262153e1acb1f5495e46e97cbcc9bac77

src/features/payments/persistence/notices.ts
Hantex9: Thanks for the suggestion! Addressed into dd831a9726d980cba6cce3f77e6187a72de689c7

LeleDallas (approved):

PR #424, opened 2024-10-17
chore: [SIW-1763] Update trial system implementation 

LazyAfternoons:
## Short description
This PR update the trial system implementation in order to align it with the production API behavior.

- Return `404` when the user is not in the trial system and when the trial id is invalid;
- Allow the initialization of the trial system to `undefined` to mimic the absence of an user in the trial system. 

## How to test
You can test this by using `io-app` and setting a mocked `trialId` like the one found in the app trial system playground.


mastro993 (approved):
LGTM

PR #425, opened 2024-10-21
[IOPID-2359] remove remote FF session refresh

Ladirico:
> [!Warning] 
> This PR depends on https://github.com/pagopa/io-services-metadata/pull/847

## Short description
Delete remote session refresh FF

## How to test
Run the application using local environment and iand using proxyman or any tool that can intercept the fetch calls made by the app you'll see that in this one here: `http://proxyman.local:3000/static_contents/status/backend.json?ms=*` the value `sessionRefresh` is not entered (as you can see from the screen)

response without sessionRefresh


Hantex9 (approved):
LGTM!

PR #430, opened 2024-10-25
feat: [IOBP-838] Remove receipt from list endpoint

LeleDallas:
## Short description
Include a summary of the changes.
This pull request includes changes to the payment route to support a new endpoint for disabling user notices.

- Add `paids/:eventId/disable` route
- Update global config with `hideReceiptResponseCode` field

### DIFF
--- a/src/features/payments/routers/notices.ts
+++ b/src/features/payments/routers/notices.ts
@@ -1,13 +1,15 @@
 import * as O from "fp-ts/lib/Option";
 import { pipe } from "fp-ts/lib/function";
-import NoticesDB from "../persistence/notices";
 import { sendFileFromRootPath } from "../../../utils/file";
+import NoticesDB from "../persistence/notices";
 
 import { NoticeListWrapResponse } from "../../../../generated/definitions/pagopa/transactions/NoticeListWrapResponse";
+import { ioDevServerConfig } from "../../../config";
 import { addNoticesHandler } from "./router";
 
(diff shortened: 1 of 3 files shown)


src/features/payments/routers/notices.ts
```diff
@@ -87,3 +89,21 @@ addNoticesHandler("get", "/paids/:eventId/pdf", (req, res) => {
...
+        pipe(
+          O.fromNullable(NoticesDB.removeUserNotice(eventId)),
+          O.fold(
+            () => res.sendStatus(hideReceiptResponseCode),
+            () => res.sendStatus(hideReceiptResponseCode)
+          )
+        )
+    )
+  );
+});
```
Hantex9, line 109: Since we don't really need two nested `fold` that have the same behavior, what do you think of replacing with the following suggestion?

```suggestion
addNoticesHandler("post", "/paids/:eventId/disable", (req, res) => {
  pipe(
    req.params.eventId,
    O.fromNullable,
    O.fold(
      () => res.sendStatus(400),
      eventId => {
        NoticesDB.removeUserNotice(eventId);
        return res.sendStatus(hideReceiptResponseCode);
      }
    )
  );
});
```

Hantex9 (changes requested):
Great work! Just a small note, and it’s good to go from my side.

Hantex9 (approved):
LGTM!

PR #433, opened 2024-11-04
[IOPID-2406] add logic trigger errorMessage and not only the errorCode in login flow

Ladirico:
## Short description
add logic trigger errorMessage and not only the errorCode in login flow

- Add new value inside the dropdown menu into the login errors
- Add logic to handle the errorMessage 

## Demo 
<p>
<video src="https://github.com/user-attachments/assets/a8683cd5-d905-431b-be2b-5ce9352da665"/>

</p>

## How to test
Run the application and checks that calls happen in the correct way and that using the SAML_RESPONSE_ERROR_MESSAGE - Missing SAMLResponse field in ACS in the drop-down menu correctly passes the errorMessage to the call (as seen in the demo)


src/routers/public.ts
```diff
@@ -95,7 +96,22 @@ addHandler(publicRouter, "get", "/idp-login", (req, res) => {
...
+      typeof req.query.error === "string" &&
+      req.query.error.includes("errorMessage:")
+    ) {
+      redirectUrl = errorMessageRedirectUrl;
+      errorCodeOrMessage = req.query.error.split(":")[1];
+    } else {
+      redirectUrl = errorCodeRedirectUrl;
+      errorCodeOrMessage = req.query.error;
+    }
+    const url = `${urlLoginScheme}://${req.headers.host}${redirectUrl}${errorCodeOrMessage}`;
```
ChrisMattew, line 114: We are already checking for `req.query.error` to be defined in the parent `if`, so this can be removed. Additionally, it's really necessary to check if `req.query.error` is a string? What do you think?

Ladirico: Screenshot 2024-11-12 alle 09 29 17
Screenshot 2024-11-12 alle 09 29 24
In order to resolve this TS errors i need to apply the logic to control if req.query.error exists and is a string

ChrisMattew (approved):
LGTM! 💪🏻

Hantex9 (approved):

PR #440, opened 2024-11-22
[IOPID-2137] Update definitions

Ladirico:
## Short description
Add `cieid` as `idp` in definitions file and add `feedback_banner_visible: true` in backend.ts (was missing value after last io-services-metadata update) 

## How to test
- Run io-app using local enviornment 
- check if the application starts and works correctly


ChrisMattew (approved):
LGTM!

PR #447, opened 2024-12-09
[IOCOM-2017] FIMS - Implicit consent for internal RP

Vangaorth:
## Short description
This PR adds support for internal Relying Parties, not requiring the user consents to exchange data since (latter) are not sent to a third party entity.

- Add a new, optional metadata for Relying Party called isInternal (Boolean). Setting isInternal to true means that the external web services referenced by the client metadata (via redirectUri) is managed by PagoPA S.p.A.
- Authentication Requests originated by internal Relying Parties no longer require explicit consent by the citizen.

## How to test
Test the authentication flow:
- isInternal: false -> will send the consents data and requires the confirm endpoint to be called
- isInternal: true -> does not send the consents and automatically follows the entire authentication flow


forrest57 (approved):
LGTM

PR #461, opened 2025-01-29
chore(CieID):[IOPID-2162] Remove CieID FF from backend.ts

ChrisMattew:
## Short description
This PR removes the optional CieID FF from the `src/payloads/backend.ts` file.


shadowsheep1 (approved):
LGTM

PR #475, opened 2025-03-26
chore: Align CDC text to IOBP-1419 requirements

LeleDallas:
## Short description
This pull request includes updates to the content and status configurations for the `Carta della Cultura` initiative enabling the CDC feature in the backend configuration.

- Update text content for `Carta della Cultura`
- Enabled the CDC feature flag

## How to test
Ensure text is aligned with [design](https://www.figma.com/design/JNcFvZKDsaEEBFQ1muc2rk/Carta-della-Cultura-2025?node-id=23-7736&t=uyEIRwjNXjwhgtxS-0)


src/payloads/backend.ts
```diff
@@ -73,7 +73,7 @@ export const backendStatus: BackendStatus = {
       opt_in_out_enabled: true
     },
     cdc: {
-      enabled: false
+      enabled: true
```
Hantex9, line 76: Using the `enabled` flag is deprecated and potentially risky, as it doesn't allow for feature rollout based on specific app versions.
I recommend refactoring the `cdc` property by removing the `enabled` flag and replacing it with the `min_app_version` property, which specifies the minimum app version required to enable the feature.

Please note that this change will also impact the `isCdcEnabledSelector` in the io-app.

Hantex9 (changes requested):

scripts/generate-api-models.sh
```diff
@@ -2,7 +2,7 @@
 
 IO_BACKEND_VERSION=v16.12.1
 # need to change after merge on io-services-metadata
-IO_SERVICES_METADATA_VERSION=1.0.66
+IO_SERVICES_METADATA_VERSION=IOBP-1419-cdc-v2
```
LeleDallas, line 5: This will be update with the newest release tag after merging this https://github.com/pagopa/io-services-metadata/pull/959

src/payloads/backend.ts
LeleDallas: Since we cannot do these changes without broke the production app, I create a `cdcV2` property that includes this functionality

scripts/generate-api-models.sh
LeleDallas: Addressed here 551117c09f8f9c9f1d771aa6c0c96ec663948d34

Hantex9 (approved):
LGTM!

PR #495, opened 2025-06-20
feature: [IOCOM-2419] Production messages in the dev-server

Vangaorth:
## Short description
This PR adds the ability to download production messages and display their content using the dev-server

- new router to download production messages `GET /downloadProductionMessages?token={bearerToken}&archived=[true|false]` where 
  - `bearerToken` is the token taken from the production login
  - `archived` is a boolean that downloads either the inbox or the archive
  - depending on the number of messages, the endpoint may take a while to respond. You must download both archived and inbox messages for the feature to work
  - messages are saved under `config/` folder, which is ignored by Git
- New parameter `useMessagesSavedUnderConfig` in `config.ts`. Set this to `true` after having downloaded both inbox and archive messages to use them
- Due to missing production services, information related to them (service name, organization name, logo, ...) are generated randomly and do not match production data
- remote content messages are not supported, nor SEND ones

## How to test
Using IO App, make sure to obtain a copy of both archived and inbox messages, then set the flag above to true, restart the dev-server and check messages on IO App.


forrest57 (approved):
LGTM!

PR #508, opened 2025-07-29
chore: [IOPID-2629] add ephemeral lollipop key manage

Ladirico:
## Short description
In this PR, the management of an ephemeral lollipop key has been added. 
When a user logs in, an ephemeral lollipop key is generated, which only becomes effective when the login is successful.
This development is used for managing active session logins.

- Create utility functions to manage the ephemeral lollipop key in src/persistence/lollipop.ts
- Added functions that manage the lollipop key within calls. In the /login call, the ephemeral key is registered, and in the /idp-login call, the ephemeral key is concretized if the login is successful; otherwise, it is deleted.

## How to test
Run the application using local enviroment and test both fast login and standard login

### DIFF
--- a/src/persistence/lollipop.ts
+++ b/src/persistence/lollipop.ts
@@ -18,6 +18,12 @@ const lollipopInfo: LollipopInfo = {
   instantiationDate: undefined
 };
 
+const lollipopInfoEhemeral: LollipopInfo = {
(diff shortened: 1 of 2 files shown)


shadowsheep1 (approved):
LGTM

PR #517, opened 2025-08-21
feat: [IOCOM-2529] SEND AAR: ToS endpoint

Vangaorth:
⚠️ This PR depends on #516 ⚠️

## Short description
This PR adds support for SEND ToS.

- New IO endpoints, `PUT /api/v1/send/tos/:consentType` to accept ToS from an AAR
- AAR endpoint (`POST /api/v1/send/aar`) has been updated to check for ToS and return 403 + data when they are not accepted

## How to test
Using an HTTP client like Postman/Bruno/cUrl, contact the IO endpoints and check that:
- calling the AAR endpoint on an AAR with no accepted ToS returns 400
- calling the new endpoint changes the ToS value (you can verify this by calling again the previous endpoint, which should return 200)
- when `tosAccepted: true` property is set in the config file (`sendAARs` section), the AAR endpoint returns 200 (for a notification that belongs the the server's user)


Alessandro-Laregina (approved):
LGTM

PR #525, opened 2025-09-04
feat: [IOCOM-2566] NIS passive SOD verifications

Vangaorth:
## Short description
This PR adds the NIS passive SOD verifications for the AAR feature.

-NIS passive SOD verifications, following the example in the "other" repository

## How to test
Either run the verifications on your CIE data or ask the developer for the test data


src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+    throw Error("Raw value found inside of ASN1 node at depth 2, child 0");
+  }
+
+  return signedDataSequence;
+};
+
+const bytesToUppercaseHex = (bytesString: string) =>
+  forge.util.bytesToHex(bytesString).toUpperCase();
+
+const consoleLogIfNeeed = (line: string, shouldLog: boolean) => {
```
Alessandro-Laregina, line 504: Just a small typo: there's an extra "e" in `consoleLogIfNeeed` — should be `consoleLogIfNeeded`.

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+    child =>
+      child.type === forge.asn1.Type.INTEGER && typeof child.value === "string"
+  );
+  if (dataGroupNode != null && typeof dataGroupNode.value === "string") {
+    return dataGroupNode.value;
+  }
+  return undefined;
+};
+
+export const dataGroupMatchingFromHashesAS1DR = (
```
Alessandro-Laregina, line 225: Looks like a small typo in the function name — `AS1DR` probably meant to be `ASN1DER`, right?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+): Either<string, void> => {
+  try {
+    const buffer = Buffer.from(sodHexString, "hex");
+    const byteStringBuffer = new forge.util.ByteStringBuffer(buffer);
+    const asn1Der = forge.asn1.fromDer(byteStringBuffer, {
+      parseAllBytes: false
+    });
+
+    const hashesOctectBlockBytes = signedDataGroupHashesFromASN1DER(asn1Der);
+    const sha256MD = forge.md.sha256.create();
```
shadowsheep1, line 21: We shouldn’t assume all digests are SHA-256. The algorithm must be derived from the corresponding OID (Object Identifier). Ex: [2.16.840.1.101.3.4.2.1](https://oid-base.com/cgi-bin/display?oid=2.16.840.1.101.3.4.2.1&submit=Go&action=display&turing-test=) -> `sha-256 (NIST Algorithm)`

What do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+      logDebugMessages
+    );
+
+    const hashedDataGroups = hashedDataGroupsFromASN1DER(asn1Der);
+    const hashesASN1Der = forge.asn1.fromDer(hashedDataGroups, {
+      parseAllBytes: false
+    });
+
+    const decodedNisStringBytes = forge.util.hexToBytes(nisInputHex);
+    const sha256MDNIS = forge.md.sha256.create();
```
shadowsheep1, line 44: As above

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+    const dataGroupNIS = dataGroupBytesFromASN1DER(foundDataGroupASN1);
+    consoleLogIfNeeded(
+      `Calculated hash (NIS) found in the SOD, DG: ${
+        dataGroupNIS != null ? bytesToUppercaseHex(dataGroupNIS) : undefined
+      }`,
+      logDebugMessages
+    );
+
+    const nisPublicKeyBytes = forge.util.hexToBytes(nisPublicKeyHex);
+    const sha256PublicKeyMD = forge.md.sha256.create();
```
shadowsheep1, line 68: As above

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+      }`,
+      logDebugMessages
+    );
+
+    const canonizedSignedAttributes = canonizeSignedAttributes(asn1Der);
+    const canonizedSignedAttributesHex = forge.asn1
+      .toDer(canonizedSignedAttributes)
+      .toHex()
+      .toUpperCase();
+    if (canonizedSignedAttributesHex !== expectedSignedAttributesHex) {
```
shadowsheep1, line 98: The expected signature attribute check was only used in the initial POC to implement the verification function. Since it must be obtained manually, I’m not sure this check should remain in the production verification function.

What do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+      );
+    }
+    consoleLogIfNeeded(
+      "Content of signed_attrs_set matches the expected value",
+      logDebugMessages
+    );
+
+    const signatureBytes = signatureFromASN1DER(asn1Der);
+    const signatureHex = bytesToUppercaseHex(signatureBytes);
+    if (signatureHex !== expectedSignatureHex) {
```
shadowsheep1, line 111: Same concern as above, but applied to `expectedSignatureHex`. What do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+    }
+    consoleLogIfNeeded(
+      "Content of signature matches the expected value",
+      logDebugMessages
+    );
+
+    const replacedSignedAttributesWrapperBytes = forge.util.hexToBytes(
+      canonizedSignedAttributesHex
+    );
+    const sha1ReplaceSignedAttributesMD = forge.md.sha1.create();
```
shadowsheep1, line 125: As above. We shouldn’t assume all digests are SHA-1. The algorithm must be derived from the corresponding OID (Object Identifier). Ex: [1.2.840.113549.1.1.5](https://oid-base.com/cgi-bin/display?oid=1.2.840.113549.1.1.5&a=display) -> `sha1WithRSAEncryption (PKCS #1)` 

What do you think?

shadowsheep1 (changes requested):
To make `verifyNIS` easier to validate, I suggest adding a stub test like this. It uses mock placeholders in the repo, but you can swap in real CIE data on the fly when running locally. This way we get test coverage structure without leaking sensitive data.

```ts
import { verifyNIS } from "../nisService";

const expectedSignatureHex =
  "mock-data";
const expectedSignedAttributesHex =
  "mock-data";
const nisInputHex = "mock-data";
const nisPublicKeyHex =
  "mock-data";
const sodHexString =
  "mock-data";
const expectedResult = false;

describe("verifyNIS", () => {
  it("should verify a valid NIS", async () => {
    const result = verifyNIS(
      expectedSignatureHex,
      expectedSignedAttributesHex,
      nisInputHex,
      nisPublicKeyHex,
      sodHexString,
      true
    );
  });
});
```

what do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+ *
+ * This function:
+ * - Takes the the `children` array out of the original `CONTEXT_SPECIFIC` wrapper (`0xA0`).
+ * - Creates a new, standard `UNIVERSAL SET` wrapper (`0x31`).
+ * - Places the original children inside this new, correct wrapper.
+ *
+ * The serialized output of the object returned by this function is the byte-for-byte correct
+ * data needed for the final hash calculation and signature verification.
+ */
+export const canonizeSignedAttributes = (
```
shadowsheep1, line 194: Several helper functions here could be extracted into a separate utilities file. That would keep the main logic cleaner and easier to follow. What do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,520 @@
...
+    }
+    const dataGroupNIS = dataGroupBytesFromASN1DER(foundDataGroupASN1);
+    consoleLogIfNeeded(
+      `Calculated hash (NIS) found in the SOD, DG: ${
+        dataGroupNIS != null ? bytesToUppercaseHex(dataGroupNIS) : undefined
+      }`,
+      logDebugMessages
+    );
+
+    const nisPublicKeyBytes = forge.util.hexToBytes(nisPublicKeyHex);
```
shadowsheep1, line 67: I suggest parsing the public key with the strict `ASN.1` parser, as we did for the `SOD` to be "trailing 0s proof". For example:

```ts
const ans1 = forge.asn1.fromDer(publicKeyDer, { parseAllBytes: false });
```

shadowsheep1 (changes requested):

shadowsheep1:
> > > > To make `verifyNIS` easier to validate, I suggest adding a stub test [...]
> > > > what do you think?
> > > 
> > > 
> > > Do we have valid public mock data that the test can run successfully?
> > > Otherwise, my idea was to later add some dedicated REST endpoints that can be called, providing the specific input validation data. Such endpoints will not be bound to the SEND AAR flow
> > 
> > 
> > Unfortunately, we don’t.
> 
> Ok so what do you think of avoiding the test and providing such dedicated endpoint, so that one can test it with a simple REST API?

Okay. Sounds good.

PR #526, opened 2025-09-05
feat: [IOCOM-2574] Non-blocking check for `isTest` parameter

Vangaorth:
## Short description
This PR adds a non-blocking check (and warning) for the query parameter `isTest=true|false` on IO AAR endpoints

Non-blocking check (with console warning) for the `isTest=true|false` parameter on the following endpoints:
- `/api/v1/send/aar`
- `/api/v1/send/notification/:iun`
- `/api/v1/send/notification/attachment/*`
- `/api/v1/send/mandate/create`
- `/api/v1/send/mandate/accept/:mandateId`

Console warninig log refactor to allow logging of a single message or an ExpressFailure type response.

## How to test
Call the endpoints above and check that:
- with no `isTest` parameter, a warning message is printed in the console
- with `isTest` parameter that has a value that is not `true` nor `false`, a warning message is printed in the console
- with `isTest=[true|false]` parameter, no warning message is logged (check for both uppercase, lowercase and mixed)

### DIFF
--- a/src/features/messages/routers/ioSendRouter.ts
+++ b/src/features/messages/routers/ioSendRouter.ts
@@ -1,4 +1,4 @@
-import { Response, Router } from "express";
+import { Request, Response, Router } from "express";
 import { lollipopMiddleware } from "../../../middleware/lollipopMiddleware";
 import { addHandler } from "../../../payloads/response";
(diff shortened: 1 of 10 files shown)


Alessandro-Laregina (approved):
LGTM

PR #528, opened 2025-09-05
feat: [IOCOM-2567] NIS verifications

Vangaorth:
## Short description
This PR adds the NIS SOD verifications for the AAR feature.

- NIS SOD verifications, following the example in the "other" repository

## How to test
Either run the verifications with your CIE data or ask the developer for the test data


shadowsheep1 (changes requested):
To make `verifyNISSOD` easier to validate, I suggest adding a stub test like this. It uses mock placeholders in the repo, but you can swap in real CIE data on the fly when running locally. This way we get test coverage structure without leaking sensitive data.

```ts
import { verifyNISSOD } from "../nisService";

const sodHexString =
  "mock-data";
const expectedResult = false;

describe("verifyNISSOD", () => {
  it("should fail to verify an invalid NIS sod chain of trust", async () => {
    const result = verifyNISSOD(sodHexString, true);
  });
});
```

what do you think?

PR #529, opened 2025-09-08
feat: [IOCOM-2568] AAR Challenge verification

Vangaorth:
## Short description
This PR adds the challenge verification, for the AAR phase 3.

- Challenge verification using the NIS public key.

## How to test
Either run the verifications with a challenge signed with your CIE data or ask the developer for the test data


shadowsheep1 (changes requested):
To make verifyChallenge easier to validate, I suggest adding a stub test like this. It uses mock placeholders in the repo, but you can swap in real CIE data on the fly when running locally. This way we get test coverage structure without leaking sensitive data.

```ts
import { verifyChallenge } from "../nisService";


const nisPublicKeyHex =
  "mock-data";
const signedChallengeHex =
  "mock-data";
const originalChallengeHex = "mock-data";
const expectedResult = false;

describe("verifyChallenge", () => {
  it("should fail to verify an invalid challenge", async () => {
    const result = verifyChallenge(
      signedChallengeHex,
      nisPublicKeyHex,
      originalChallengeHex,
      true
    );
  });
});

```

What do you think?

src/features/pn/services/cie/nisService.ts
```diff
@@ -0,0 +1,736 @@
...
+): Either<string, void> => {
+  try {
+    // --- 1. Prepare the Public Key ---
+    // The public key is provided as a hex string representing its DER format.
+    // It must be converted into a node-forge public key object to be used.
+
+    // Convert the hex-encoded public key into a binary string.
+    const publicKeyDer = forge.util.hexToBytes(publicKeyHex);
+    // Parse the binary DER string into an ASN.1 object structure.
+    const ans1 = forge.asn1.fromDer(publicKeyDer);
```
shadowsheep1, line 292: I suggest parsing the public key with the strict ASN.1 parser, as we did for the SOD to be "trailing 0 bullet proof". For example:


```suggestion
    const asn1 = forge.asn1.fromDer(publicKeyDer, { parseAllBytes: false });
```

shadowsheep1 (changes requested):

PR #531, opened 2025-09-09
fix: [IOCOM-2585] Fix for `x-pagopa-pn-io-src` header and 403 on AAR QRCode not belonging to user

Vangaorth:
## Short description
This PR fixes the `x-pagopa-pn-io-src` header and a bad response code for the QRCode API.

- `x-pagopa-pn-io-src` header's value was not being transferred from client to SEND router
- 403 is returned when an AAR QRCode not belonging to the user is requested

## How to test
Request a notification (IO Endpoint) without the `x-pagopa-pn-io-src` header set: a warning should appear in the console. Check it again with the value of such header set to "QRCODE": no warning should appear
Check the AAR endpoint (IO Endpoint) with a QRCode content that does not belong to the user: it should return 403.

### DIFF
--- a/src/features/messages/services/ioSendService.ts
+++ b/src/features/messages/services/ioSendService.ts
@@ -22,9 +22,21 @@ export const generateRequestHeaders = (
   ...MessagesService.sendTaxIdHeader(
     ioDevServerConfig.profile.attrs.fiscal_code
   ),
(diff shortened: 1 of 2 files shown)


forrest57 (approved):
LGTM, tested both attachment and notifications endpoints

PR #581, opened 2026-03-13
feat: [IEL-185] Fci flow with generated nonce

soixdev91:
## Short description
This PR introduces nonce lifecycle handling for the FCI sign flow, allowing to test the IO app behaviour with expired nonce running it with local configuration.

- Updated GET `/api/v1/sign/qtsp/clauses` to generate and return a fresh nonce, storing its expiration.
- Updated POST `/api/v1/sign/signatures` to validate `qtsp_clauses.nonce` and return an error (500) when nonce is not valid.
- Extended FCI router tests to cover nonce validation.

## How to test
Run the IO app with .env.local, then start the FCI signing flow (the dev-server config must include at least one fci.waitForSignatureCount). Stop at any step before final signing. After the nonce validity window (fci.response.nonceDuration) has expired, try to complete the signature: POST /api/v1/sign/signatures should return 500, and the app should show an error message with a button to restart the signing flow.

### DIFF
--- a/src/features/fci/qtspNonceStore.ts
+++ b/src/features/fci/qtspNonceStore.ts
@@ -2,7 +2,7 @@ import { randomUUID } from "crypto";
 import { ioDevServerConfig } from "../../config";
 
 export const QTSP_NONCE_EXPIRING_MS =
-  ioDevServerConfig.messages.fci.response.nonceDuration * 1000;
+  ioDevServerConfig.messages.fci.response.nonceDurationSeconds * 1000;
 
 const qtspNonceExpirations = new Map<string, Date>();
 
--- a/src/features/fci/qtspNonceStore.ts
+++ b/src/features/fci/qtspNonceStore.ts
@@ -1,0 +1,56 @@
+import { randomUUID } from "crypto";
+import { ioDevServerConfig } from "../../config";
+
+export const QTSP_NONCE_EXPIRING_MS =
+  ioDevServerConfig.messages.fci.response.nonceDuration * 1000;
+
+const qtspNonceExpirations = new Map<string, Date>();
+
+const cleanupExpiredQtspNonces = (referenceDate: Date) => {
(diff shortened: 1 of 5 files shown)


Ladirico (changes requested):

src/features/fci/qtspNonceStore.ts
```diff
@@ -0,0 +1,56 @@
...
+  cleanupExpiredQtspNonces(now);
+  const nonce = generateQtspNonce();
+  qtspNonceExpirations.set(
+    nonce,
+    new Date(now.getTime() + QTSP_NONCE_EXPIRING_MS)
+  );
+  return nonce;
+};
+
+export type QtspNonceValidationResult = "valid" | "expired" | "missing";
```
freddi301: why do we need to disambiguate between different validation results? isn't a boolean enough? i see in the diff that only === "valid" checks are done

src/features/fci/qtspNonceStore.ts
```diff
@@ -0,0 +1,56 @@
+import { randomUUID } from "crypto";
+import { ioDevServerConfig } from "../../config";
+
+export const QTSP_NONCE_EXPIRING_MS =
+  ioDevServerConfig.messages.fci.response.nonceDurationSeconds * 1000;
+
+const qtspNonceExpirations = new Map<string, Date>();
+
+const cleanupExpiredQtspNonces = (referenceDate: Date) => {
```
freddi301: do we need to cleanup nonces at all? do we know how the backend manages this? after how much time the actual backend deltes expired nonces?

src/features/fci/qtspNonceStore.ts
soixdev91: addressed in 73234962c61bef0777c2daaf31fed73868f5d88e

freddi301 (approved):
LGTM

Ladirico (approved):

PR #599, opened 2026-05-07
chore: [IOPID-3902] Add bearer authorization in services APIs and fixes SEND activation route

Vangaorth:
## Short description
This PR adds the bearer authorization header to the services APIs and fixes the new SEND activation route 

- New specs for services' APIs generation
- Removed bad new url for SEND special service activation, replaced with the old logic for the new endpoint
- Common code shared for the installation endpoint

## How to test
Using the related branch of IO-App, check that the services section loads fine and that the SEND special service can be properly turned on and off.


soixdev91 (approved):
Send activation now works as expected!

adelloste (approved):
LGTM

PR #618, opened 2026-08-07
refactor: replace fp-ts Option with plain conditional in start.ts

Leerti:
## Short description
Replaced fp-ts `Option`/`pipe` usage with a plain null check for formatting `row.description`, removing an unnecessary fp-ts dependency for a simple conditional.

- Replaced fp-ts `Option`/`pipe` usage with a plain null check for formatting `row.description`
- Removes unnecessary fp-ts dependency for a simple conditional
-
## How to test
- Run `yarn start` and verify the routes table prints descriptions correctly, both for routes with a description and without one

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -2,8 +2,6 @@ import child_process from "child_process";
 import chalk from "chalk";
 import { cli } from "cli-ux";
 import figlet from "figlet";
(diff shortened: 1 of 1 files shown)


Vangaorth (approved):
LGTM

